PR #18, opened 2018-08-22
docs: add quick start docs

danbf:
docs: add quick start docs

### DIFF
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -19,7 +19,7 @@ test -f config/proxy_configs.yml || die "Error: could not locate \`config/proxy_
 Copy and configure from \`config/proxy_configs.yml.example\`"
 
 # lets find out the host ip and then use that to connect the sso-proxy to the nginx-proxy
-export host_ip=`docker run buzzfeed/cop:fix-dockerfile ping -c 1 host.docker.internal | grep icmp_seq | awk '{print $4}' | cut -d':' -f1`
+export host_ip=`docker run buzzfeed/sso:latest ping -c 1 host.docker.internal | grep icmp_seq | awk '{print $4}' | cut -d':' -f1`
 
 #create cookies if they don't exist otherwise use the ones already created
 test -f proxy_cookie_secret || openssl rand -base64 32 | head -c 32 | base64 > proxy_cookie_secret


docs/overview.md
```diff
@@ -0,0 +1,35 @@
+# Overview
```
mccutchen: This seems to duplicate the README.  Why do we need both?  Is this linked from somewhere else?

docs/quickstart.md
```diff
@@ -0,0 +1,176 @@
+# Initial Setup / Operations
+
+`SSO` is used to provide single-sign-on authentication and authorization for internal web
+applications, however, there are a few one-time prerequisites that need to be handled manually.
```
mccutchen: This sentence is a bit odd, the first clause and second clause seem like the should be related because of the "however," but I think they are totally unrelated.  By the time a person has reached this doc, I think they know what sso is.

I think we can just drop it altogether, as it doesn't provide much context that we're not already providing immediately below at the beginning of the "Google OAuth Setup" step!

docs/quickstart.md
```diff
@@ -0,0 +1,176 @@
...
+
+**NOTE: all commands should be performed on your HOST OS.**
```
mccutchen: I think this is another holdover from the rig docs?  Let's drop it!

docs/quickstart.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Google OAuth Setup
+
+Configuring `sso` requires that we first configure a Google OAuth Project.
```
mccutchen: Here is where I might add a note that these steps (unfortunately) can't be automated.

docs/quickstart.md
```diff
@@ -0,0 +1,176 @@
...
+will need to provide the following "API Scopes" for the service account that we have created:
+- `https://www.googleapis.com/auth/admin.directory.group.readonly`
+- `https://www.googleapis.com/auth/admin.directory.user.readonly`
+
+Type these into the field titled "One or More API Scopes", as a comma-separated list. After these
+have been filled in, click the "Authorize" button.
+
+#### Configuring `sso` to act as an administrator
+
+When running the `sso-authenticator` binary, the following environment variables must be set, in
```
mccutchen: `sso-auth`

docs/quickstart.md
```diff
@@ -0,0 +1,176 @@
...
+order to act as an administrator and access user group membership data:
+- **`BUZZFEED_GOOGLE_ADMIN_EMAIL`**: An administrative email address on your organization's
+domain, the identity of which can be assumed by `sso`.
+- **`BUZZFEED_GOOGLE_SERVICE_ACCOUNT_JSON`**: A permanent location for the private key `.json` file 
+that was downloaded through your browser at the time of service account creation. There is no
+reason why this file should ever be accessed by any person or service other than `sso`; ensure that
+file permissions are set accordingly.
+
+If you already have an instance of `sso` running, then it will need to be restarted before these
+changes will take effect.
```
mccutchen: I'm not sure this note is useful in the context of a quickstart doc!

docs/quickstart.md
```diff
@@ -0,0 +1,176 @@
...
+
+
+## Run the containers
+
+Four containers are in the quickstart docker-compose.yml included in this repo. They are `sso_auth` and `sso_proxy`
+containers, which run the `sso` code. A `nginx-proxy` container which acts as a dynamic proxy front-ending the cluster,
+and lastly a `payload` container which runs the payload that's `sso` is providing auth services for.
+
+Copy the `docker-compose.yml.example`, `env.example` and `proxy_configs.yml.example` files to new instances removing the
+.example trailer on the file name.
```
mccutchen: Can we simplify this?  Ideally, I think these instructions should reduce down to:

```
1. Copy your google service account JSON to `google-service-account.json`
2. Copy env.example to `.env` and fill in the blanks with the Google OAuth credentials created above
3. Run `docker-compose up -d`
```

Maybe we need two different env files in step 2, one for each service?

docs/quickstart.md
```diff
@@ -0,0 +1,176 @@
...
+Configure SSO
+ 
+<flesh this out>
+
+
+## Run the containers
+
+Four containers are in the quickstart docker-compose.yml included in this repo. They are `sso_auth` and `sso_proxy`
+containers, which run the `sso` code. A `nginx-proxy` container which acts as a dynamic proxy front-ending the cluster,
+and lastly a `payload` container which runs the payload that's `sso` is providing auth services for.
```
mccutchen: I think this is too much detail for a quickstart.  All of this information can be gleaned from the docker-compose.yml file, right?

If we do want to provide this explanatory context, let's move it down below, _after_ the user has hopefully gotten the services up and running using `docker-compose`!

danbf:
this still needs a bit of love in the `Configure SSO` section, i.e. how/where to put the all the auth stuff google provides into the quickstart templates to get them to work right, also to setup the host side `/etc/hosts` file entries

```
127.0.0.1 payload.sso.sso-test.io
127.0.0.1 sso-auth.sso-test.io
```

mccutchen:
> also to setup the host side /etc/hosts file entries

@danbf the easiest thing I can think of is to do this in the `quickstart` script!  Like, check for the expected entries, and if they're not found tell the user we're adding them before doing some `sudo sh -c 'echo "blah" >>/etc/hosts'` trick that will prompt them for their password.

What do you think?

docs/quickstart.md
```diff
@@ -0,0 +1,176 @@
...
+credentials for the service. Click the button labeled "Create credentials", and select the option
+"OAuth client ID".
+
+![Create Credentials Dropdown](images/setup-create_credentials.jpg)
+
+The next screen will ask you to choose an "Application Type"; select "Web application". You will be
+asked to fill in three fields.
+- **Name**: Any appropriate name is fine (e.g. `Dev`).
+- **Authorized JavaScript origins**: We recommend leaving this field blank.
+- **Authorized redirect URIs**: Add the URI of your `sso-proxy` deployment, with the path suffix
```
mccutchen: I'm pretty sure this needs to point at the _sso-auth_ deployment, not the sso-proxy deployment.

docker-compose.yml.example
```diff
@@ -0,0 +1,84 @@
...
+services:
+  nginx-proxy:
+    image: jwilder/nginx-proxy:alpine
+    ports:
+      - "80:80"
+    volumes:
+      - /var/run/docker.sock:/tmp/docker.sock:ro
+
+
+  proxy:
```
mccutchen: Let's call this `sso-proxy` for consistency!

docker-compose.yml.example
```diff
@@ -0,0 +1,84 @@
...
+      - CLIENT_SECRET=
+      - AUTH_CODE_SECRET=
+      - HTTP_PUBLIC_CNAME=sso-proxy.sso.sso-test.io
+      - HOST=sso-proxy.sso.sso-test.io
+      - CNAME=sso-proxy.sso.sso-test.io
+      - COOKIE_SECRET=${proxy_cookie_secret}
+      - OLD_COOKIE_SECRET=${old_cookie_secret}
+
+
+  authenticator:
```
mccutchen: Likewise, let's call this `sso-auth`!

danbf:
good point in https://github.com/buzzfeed/sso/pull/18#issuecomment-415206570 .  that's how i've been avoiding having to document manual steps like the cookie generation, just automate them.

docker-compose.yml.example
danbf: good point, fixed

docker-compose.yml.example
danbf: also good point, fixed.

docs/overview.md
danbf: was wrestling with this myself, it's out of scope for the quickstart so i'll ditch it.

docs/quickstart.md
danbf: ditching

docs/quickstart.md
danbf: correct, the location of the quickstart is `http://sso-auth.sso-test.io/oauth2/callback`, fixing.

docs/quickstart.md
danbf: fixed

docs/quickstart.md
danbf: yeah, have a bit more cleanup to do here.

Issue #20, opened 2018-08-22
switch to using alpine or a smaller container


danbf:
switch to using alpine or a smaller container, like golang:1.9-alpine

aiman-alsari:
Just a note, I built the image using alpine and couldn't replicate the issue described in #22:

> Apparently something about the alpine image prevents the pure Go DNS resolver from properly reading /etc/hosts, which breaks the quick start:
> https://groups.google.com/d/msg/golang-nuts/G-faJ0bthz0/Lx-MsDIOCAAJ

Is it possible this has been fixed by an alpine update, or is there some specific code path I'm not testing?
@mccutchen any thoughts?

PR #22, opened 2018-08-22
docker: switch to multistage build

mccutchen:
This results in a minimal final image without the full golang runtime, addressing #20.

RFR @loganmeetsworld @shrayolacrayon 
cc @danbf

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -1,7 +1,6 @@
 github.com/18F/hmacauth                  1.0.1
 gopkg.in/yaml.v2                         v2
 github.com/imdario/mergo                 v0.3.4
-github.com/BurntSushi/toml               d94612f9fc140360834f9742158c70b5c5b5535b
 github.com/bitly/go-simplejson           da1a8928f709389522c8023062a3739f3b4af419
 github.com/mreiferson/go-options         77551d20752b54535462404ad9d877ebdb26e53d
 github.com/datadog/datadog-go/statsd	 c74bd0589c83817c93e4eff39ccae69d6c46df9b
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,32 @@
-FROM golang:1.9.2
+# =============================================================================
+# build stage
+#
+# install golang dependencies & build binaries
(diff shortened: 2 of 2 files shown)


Godeps
```diff
@@ -1,7 +1,6 @@
 github.com/18F/hmacauth                  1.0.1
 gopkg.in/yaml.v2                         v2
 github.com/imdario/mergo                 v0.3.4
-github.com/BurntSushi/toml               d94612f9fc140360834f9742158c70b5c5b5535b
```
loganmeetsworld: Thought we'd already removed this!

mccutchen:
Size difference, for posterity:

```
$ docker image list | grep sso
sso                             master              57eb2a5f0f46        About a minute ago       1.13GB
sso                             alpine              cc20033f631e        10 minutes ago           26.3MB
```

Went from 1.13GB to 26.3MB!

mccutchen:
Check latest commit, we need to add the `ca-certificates` package to the final image.  Without it, we get an error trying to talk to Google:

```
{"error":"Post https://www.googleapis.com/oauth2/v3/token: x509: failed to load system roots and no roots provided","level":"error","msg":"error redeeming authentication code","remote_address":"172.20.0.1","service":"sso-auth","time":"2018-08-23 00:57:22.82312"}
```

mccutchen:
Another update:  Apparently something about the alpine image prevents the pure Go DNS resolver from properly reading `/etc/hosts`, which breaks the quick start:
https://groups.google.com/d/msg/golang-nuts/G-faJ0bthz0/Lx-MsDIOCAAJ

Now we've updated to `debian:stable-slim` instead of `alpine:3.8` for the base image, bringing our final image size up to 83MB:

```
sso                             debian              888f18f0bcc5        39 seconds ago      83.5MB
```

danbf:
wow, great size savings!

mccutchen:
Yep! Just wanted to makes sure y’all had a chance to see the individual
fixes before I squashed them away!

PR #48, opened 2018-09-13
Don't check in the dist/ directory created by make

sporkmonger:
Currently the `dist/` directory created by `make` isn't in `.gitignore` but it probably should be.


sporkmonger:
Ready for review. I think this one is simple enough to not require tests. 😁

PR #49, opened 2018-09-15
Make TLS verification of upstream servers configurable

### LINKED ISSUE #47: Add TLS verification configuration option to proxy
Currently, the proxy verifies TLS certificates are valid before sending requests to upstream servers. We'd like to point directly at AWS ALB DNS records in some cases, which means those certs are essentially guaranteed not to be valid. The validity of the cert buys very little in our threat model since it's not mutual auth, it's trivial to get a valid cert, and it's the firewall/security group rules that are what's really giving you any kind of assurance.

I'd like to either have a TLS_VERIFY environment variable that controls whether the proxy will verify certificates or not, or add a config line to the upstream_configs.yml file that would control it. The latter being preferable since conceivably you might have some upstream services that do have valid certs and some that don't.

sporkmonger:
When pointing the proxy directly at an AWS ALB, e.g., the cert presented by the ALB is guaranteed to be invalid, as you can't get a cert for `*.elb.amazonaws.com`. Depending on your deployment setup you may or may not be able to easily automate creation of corresponding DNS records to match your certs. And of course your certs might be self-signed. Therefore it's desirable to be able to disable TLS verification for upstream servers in some cases. Particularly because the real security barrier is the necessary firewall rule that ensures nothing else can talk to the upstream server.

Fixes #47.

### DIFF
--- a/internal/proxy/oauthproxy.go
+++ b/internal/proxy/oauthproxy.go
@@ -1,6 +1,7 @@
 package proxy
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -131,11 +132,13 @@ func (u *UpstreamProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // upstreamTransport is used to ensure that upstreams cannot override the
 // security headers applied by sso_proxy
-type upstreamTransport struct{}
+type upstreamTransport struct {
+	transport *http.Transport
+}
 
 // RoundTrip round trips the request and deletes security headers before returning the response.
 func (t *upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	resp, err := http.DefaultTransport.RoundTrip(req)
+	resp, err := t.transport.RoundTrip(req)
 	if err != nil {
 		logger := log.NewLogEntry()
 		logger.Error(err, "error in upstreamTransport RoundTrip")
@@ -147,11 +150,29 @@ func (t *upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error)
 	return resp, err
 }
 
+func newUpstreamTransport(insecureSkipVerify bool) *upstreamTransport {
+	return &upstreamTransport{
+		transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+				DualStack: true,
+			}).DialContext,
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			TLSClientConfig:       &tls.Config{InsecureSkipVerify: insecureSkipVerify},
+			ExpectContinueTimeout: 1 * time.Second,
+		},
+	}
+}
+
 // NewReverseProxy creates a reverse proxy to a specified url.
 // It adds an X-Forwarded-Host header that is the request's host.
-func NewReverseProxy(to *url.URL) *httputil.ReverseProxy {
+func NewReverseProxy(to *url.URL, config *UpstreamConfig) *httputil.ReverseProxy {
 	proxy := httputil.NewSingleHostReverseProxy(to)
-	proxy.Transport = &upstreamTransport{}
+	proxy.Transport = newUpstreamTransport(config.TLSSkipVerify)
 	director := proxy.Director
 	proxy.Director = func(req *http.Request) {
 		req.Header.Add("X-Forwarded-Host", req.Host)
@@ -164,9 +185,9 @@ func NewReverseProxy(to *url.URL) *httputil.ReverseProxy {
 // NewRewriteReverseProxy creates a reverse proxy that is capable of creating upstream
 // urls on the fly based on a from regex and a templated to field.
 // It adds an X-Forwarded-Host header to the the upstream's request.
-func NewRewriteReverseProxy(route *RewriteRoute) *httputil.ReverseProxy {
+func NewRewriteReverseProxy(route *RewriteRoute, config *UpstreamConfig) *httputil.ReverseProxy {
 	proxy := &httputil.ReverseProxy{}
-	proxy.Transport = &upstreamTransport{}
+	proxy.Transport = newUpstreamTransport(config.TLSSkipVerify)
 	proxy.Director = func(req *http.Request) {
 		// we do this to rewrite requests
 		rewritten := route.FromRegex.ReplaceAllString(req.Host, route.ToTemplate.Opaque)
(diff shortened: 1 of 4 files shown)


internal/proxy/proxy_config.go
```diff
@@ -48,6 +48,7 @@ type UpstreamConfig struct {
 
 	SkipAuthCompiledRegex []*regexp.Regexp
 	AllowedGroups         []string
+	TLSVerify             bool
```
shrayolacrayon: It looks like in this case, the default for `TLSVerify` would be false,  which would be a less secure default. Can you change this field to be `InsecureSkipVerify`, which when set to true, will enable `InsecureSkipVerify`? This will allow us to default to a more secure configuration.

internal/proxy/proxy_config.go
```diff
@@ -75,6 +76,7 @@ type OptionsConfig struct {
 	HeaderOverrides map[string]string `yaml:"header_overrides"`
 	SkipAuthRegex   []string          `yaml:"skip_auth_regex"`
 	AllowedGroups   []string          `yaml:"allowed_groups"`
+	TLSVerify       bool              `yaml:"tls_verify"`
```
shrayolacrayon: (nit) Can we add a unit test for this config value to ensure that it is being set?

internal/proxy/proxy_config.go
sporkmonger: Good point, and agree that default should be to verify. I meant to address that and forgot. However `InsecureSkipVerify` is a good name only inside a config struct named `TLSConfig`. It's not obvious that it's even TLS related outside that context. Given that, I'd be inclined to explicitly set a default rather than relying on zero values.

internal/proxy/proxy_config.go
sporkmonger: Absolutely. I was having trouble understanding your test suite when I first wrote the PR, but a couple PRs in, I think it should be no problem.

sporkmonger:
Added tests for TLS verification, wanted to make sure everyone's OK w/ keeping the `TLSVerify` configuration key name and just explicitly setting a default value before proceeding.

internal/proxy/proxy_config.go
shrayolacrayon: What if we make the name `TLSSkipVerify`?

sporkmonger: Main reason I'd choose `TLSVerify` over `TLSSkipVerify` is that `TLSVerify` / `TLS_VERIFY` and similar are in wide usage in other tools, whereas `TLSSkipVerify` would be relatively uncommon. I also generally don't like to let language details (like what Go's zero values are) to leak out into user-facing interface design, and I think of config as part of the interface design.

sporkmonger: That said, there's certainly other tools using `TLSSkipVerify`, like Vault and KSonnet.

sporkmonger:
Ended up writing the code for it because it turned out to be easier than I thought it would be, but it's super easy to switch if you do prefer `TLSSkipVerify`.

internal/proxy/proxy_config.go
```diff
@@ -360,7 +363,9 @@ func parseOptionsConfig(proxy *UpstreamConfig) error {
 	proxy.Timeout = proxy.RouteConfig.Options.Timeout
 	proxy.FlushInterval = proxy.RouteConfig.Options.FlushInterval
 	proxy.HeaderOverrides = proxy.RouteConfig.Options.HeaderOverrides
-	proxy.TLSVerify = proxy.RouteConfig.Options.TLSVerify
+	if proxy.RouteConfig.Options.TLSVerify != nil {
+		proxy.TLSVerify = *proxy.RouteConfig.Options.TLSVerify
+	}
```
sporkmonger: One item of note, we don't want to set the explicit default value here because of the `if proxy.RouteConfig.Options == nil` short-circuit at the top of the function. Doing so causes half the test suite to fail.

internal/proxy/proxy_config.go
```diff
@@ -75,6 +76,7 @@ type OptionsConfig struct {
 	HeaderOverrides map[string]string `yaml:"header_overrides"`
 	SkipAuthRegex   []string          `yaml:"skip_auth_regex"`
 	AllowedGroups   []string          `yaml:"allowed_groups"`
+	TLSVerify       *bool             `yaml:"tls_verify"`
```
shrayolacrayon: I think it still makes sense to use `TLSSkipVerify` as a `bool`, which removes the need for any default setting on load of the service configs.

sporkmonger: OK, I'll change it.

sporkmonger:
OK, it's been switched to `TLSSkipVerify`.

internal/proxy/oauthproxy_test.go
```diff
@@ -201,6 +201,110 @@ func TestNewReverseProxyHostname(t *testing.T) {
 
 }
 
+func TestNewReverseProxyTLSSkipVerifyFalse(t *testing.T) {
```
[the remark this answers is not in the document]
sporkmonger: I prefer table driven tests, was just mirroring what was already there. Happy to convert over.

sporkmonger:
Does it make it easier to review if I squash earlier in the PR or at the end?

shrayolacrayon:
At the end would be easier, thanks for asking, this also looks good to me if you'd like to rebase + squash, we can get this merged!

sporkmonger:
All rebased and squashed!

internal/proxy/oauthproxy.go
```diff
@@ -131,11 +132,26 @@ func (u *UpstreamProxy) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 // security headers applied by sso_proxy
-type upstreamTransport struct{}
+type upstreamTransport struct {
+	InsecureSkipVerify bool
+}
 
 // RoundTrip round trips the request and deletes security headers before returning the response.
 func (t *upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	resp, err := http.DefaultTransport.RoundTrip(req)
+	transport := &http.Transport{
```
shrayolacrayon: Rather than creating a new transport every time, can we make this initialize one once and reuse it instead? This will hopefully reduce the overhead of TCP connections made.

shrayolacrayon:
Added one more request!

sporkmonger:
No worries! Downside is it's a little more repetitious this way, but because of the shape of the structure and pointers, it's a little more challenging to factor that out in a way that wouldn't be buggy. Open to ideas.

internal/proxy/oauthproxy.go
```diff
@@ -167,7 +154,21 @@ func (t *upstreamTransport) RoundTrip(req *http.Request) (*http.Response, error)
 // It adds an X-Forwarded-Host header that is the request's host.
 func NewReverseProxy(to *url.URL, config *UpstreamConfig) *httputil.ReverseProxy {
 	proxy := httputil.NewSingleHostReverseProxy(to)
-	proxy.Transport = &upstreamTransport{InsecureSkipVerify: config.TLSSkipVerify}
+	proxy.Transport = &upstreamTransport{
```
shrayolacrayon: nit: we can probably create a function that initializes the `upstreamTransport{}`, something like `NewUpstreamTransport(insecureSkipVerify bool)` so that we're not repeating the config code twice?

sporkmonger: Yeah, I like that solution better.

sporkmonger:
OK, I think this should be ready to go. Squash again?

while1eq1:
Beautiful. Ive already been using this prior to the merge. Thanks @sporkmonger

PR #51, opened 2018-09-17
internal/pkg/aead: use base64.URLRawEncoding

shrayolacrayon:
## Problem
Sporadically, we are getting a `"cipher: message authentication failed"` error which might be related to the additional padding added to the encrypted session state, that comes with using `base64.URLEncoding`.

## Solution
Use `base64.RawURLEncoding`, which omits padding characters when encoding the encrypted session state.

### DIFF
--- a/internal/pkg/aead/aead.go
+++ b/internal/pkg/aead/aead.go
@@ -98,7 +98,7 @@ func (c *MiscreantCipher) Marshal(s interface{}) (string, error) {
 	}
 
 	// base64-encode the result
-	encoded := base64.URLEncoding.EncodeToString(ciphertext)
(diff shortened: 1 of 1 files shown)


jphines:
This overflow post, https://stackoverflow.com/questions/47382035/unable-to-decrypt-after-aes-gcm-base64-in-go, suggests the issue here is that the output encoding adds padding such that the resulting output value can't be decrypted successfully.

Intuitively, this makes sense. We've also verified the API we're currently using, namely `base64.URLEncoding` does add padding which could be problematic, from https://golang.org/pkg/encoding/base64/#NewEncoding

which is used to instantiate
```golang
var URLEncoding = NewEncoding(encodeURL)
```

Using a pad-less solution here seems like it should fix this problem. 💯 

LGTM.

Issue #42, opened 2018-08-28
sso-proxy: remove deprecated session unmarshal code and old cookie secrets


shrayolacrayon:
**Why**
We have two code paths for loading the session state in `sso-proxy` that we kept in until the cookie refresh period expired for sso in production. Both code paths were kept in place so that having an invalidly decrypted session cookie would not affect the UX for those logging into sso-proxy with the sessions encrypted the old way.  We can now remove that code path. 

This will involve removing:
- `OldCookieCipher` and `OldCookieSecret` from the `OAuthProxy` [struct](https://github.com/buzzfeed/sso/blob/master/internal/proxy/oauthproxy.go#L59). 
- [initialization logic] for old cookie secrets and ciphers (https://github.com/buzzfeed/sso/blob/master/internal/proxy/oauthproxy.go#L249) in `NewOAuthProxy`
- [`ErrRefreshCookie`](https://github.com/buzzfeed/sso/blob/master/internal/proxy/oauthproxy.go#L921)
- fallback of using old cookie cipher  in [`LoadCookiedSession`](https://github.com/buzzfeed/sso/blob/master/internal/proxy/oauthproxy.go#L491)

loganmeetsworld:
Handled in https://github.com/buzzfeed/sso/pull/54

Issue #41, opened 2018-08-28
sso-auth: remove deprecated payloads package and alternate code path for loading session


shrayolacrayon:
**Why**
We have two code paths for loading the session state in `sso-auth` that we kept in until the cookie refresh period expired for sso in production. We can now remove that code as well as all of the payloads package, and only use the miscreant implementation of AES-SIV for encrypting and decrypting session state. 

This would involve removing:
- [`AuthLoadSession`](https://github.com/buzzfeed/sso/blob/master/internal/pkg/sessions/cookie_store.go#L193)
- [`setPayloads`](https://github.com/buzzfeed/sso/blob/master/internal/auth/authenticator.go#L90)
- [`internal/pkg/payloads`](https://github.com/buzzfeed/sso/tree/master/internal/pkg/payloads)

Issue #29, opened 2018-08-24
sso-auth: segfault if `OLD_COOKIE_SECRET` value is not present


danbf:
reproduce by running the quick-start docker-compose environment without the OLD_COOKIE_SECRET causes a seg-fault

```
authenticator_1 | panic: runtime error: invalid memory address or nil pointer dereference
authenticator_1 | [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x6ba2d6]
authenticator_1 |
authenticator_1 | goroutine 5 [running]:
authenticator_1 | github.com/buzzfeed/cop/internal/pkg/sessions.(*CookieStore).AuthLoadSession(0x0, 0xc420127500, 0x1, 0x10, 0xc420074c30)
authenticator_1 | /go/src/github.com/buzzfeed/cop/internal/pkg/sessions/cookie_store.go:196 +0x26
authenticator_1 | github.com/buzzfeed/cop/internal/auth.(*Authenticator).LoadSession(0xc4200b6500, 0xc420127500, 0xa, 0x7fdc319a7000, 0x0)
authenticator_1 | /go/src/github.com/buzzfeed/cop/internal/auth/authenticator.go:144 +0xa6
authenticator_1 | github.com/buzzfeed/cop/internal/auth.(*Authenticator).authenticate(0xc4200b6500, 0xadb360, 0xc420176000, 0xc420127500, 0x1, 0xc420018270, 0x22)
authenticator_1 | /go/src/github.com/buzzfeed/cop/internal/auth/authenticator.go:283 +0x80
authenticator_1 | github.com/buzzfeed/cop/internal/auth.(*Authenticator).SignIn(0xc4200b6500, 0xadb360, 0xc420176000, 0xc420127500)
authenticator_1 | /go/src/github.com/buzzfeed/cop/internal/auth/authenticator.go:372 +0x14c
authenticator_1 | github.com/buzzfeed/cop/internal/auth.(*Authenticator).SignIn-fm(0xadb360, 0xc420176000, 0xc420127500)
authenticator_1 | /go/src/github.com/buzzfeed/cop/internal/auth/authenticator.go:207 +0x48
authenticator_1 | github.com/buzzfeed/cop/internal/auth.(*Authenticator).validateSignature.func1(0xadb360, 0xc420176000, 0xc420127500)
authenticator_1 | /go/src/github.com/buzzfeed/cop/internal/auth/middleware.go:154 +0x121
[...]

PR #54, opened 2018-09-17
sso: remove old cipher fallback

jphines:
## Problem

There is a legacy fallback cipher that was used when we migrated session encryption schemas. This legacy fallback is no longer used and adds needless complexity.

## Solution

We should remove the legacy fallback cipher.

## Notes

This removes the legacy fallback cipher in both sso proxy and authenticator.

Fixes https://github.com/buzzfeed/sso/issues/41
Fixes https://github.com/buzzfeed/sso/issues/42


jphines:
Ready For Review @shrayolacrayon 
cc/ @buzzfeed/infra-security

shrayolacrayon:
This looks great! ✂️ 
You can also remove `internal/pkg/payloads` if it's not already gone.

jphines:
@shrayolacrayon removed those pkgs

shrayolacrayon:
lgtm!

sporkmonger:
Did this fix #29? Can that be closed now?

PR #55, opened 2018-09-18
sso-proxy: add preserve host configuration option

### LINKED ISSUE #52: Support a configuration option akin to mod_proxy's ProxyPreserveHost
Currently SS Octopus replaces the Host header provided by the client with a new Host header based on the upstream DNS name. It also isn't very careful in how it sets the `X-Forwarded-Host` header (adds a new one, ignoring any that may have been set by another proxy or networking component). In some configurations, particularly those involving other proxies or servers that have virtual hosts, it's desirable to preserve the client-provided host. This should be a per-upstream configuration option.

sporkmonger:
When pointing the proxy directly at an AWS ALB, e.g., the host name will be `xxx.yyy.zzz.elb.amazonaws.com`, however the service itself will typically be expecting a hostname akin to `servicename.intranet.company.com`. This change adds a config option that will send the original `Host` header sent by the client through without change. It also changes the behavior of the `X-Forwarded-Host` header handling code so that the header will only be set if it doesn't already exist. Currently the existing code will create a second `X-Forwarded-Host` with potentially a different value if there's a proxy chain, which seems likely to be a bug.

Since the new config option is named `PreserveHost` and the default zero value is `false`, the default behavior should continue to be the current behavior. However for either config value this will fix the broken `X-Forwarded-Host` behavior.

Fixes #52. Will conflict with #49. I'll clean up whichever gets merged second.

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -1,8 +1,6 @@
 github.com/18F/hmacauth                  1.0.1
 gopkg.in/yaml.v2                         v2
 github.com/imdario/mergo                 v0.3.4
-github.com/bitly/go-simplejson           da1a8928f709389522c8023062a3739f3b4af419
-github.com/mreiferson/go-options         77551d20752b54535462404ad9d877ebdb26e53d
 github.com/datadog/datadog-go/statsd	 c74bd0589c83817c93e4eff39ccae69d6c46df9b
 golang.org/x/oauth2                      7fdf09982454086d5570c7db3e11f360194830ca
 golang.org/x/net/context                 242b6b35177ec3909636b6cf6a47e8c2c6324b5d
--- a/internal/proxy/options.go
+++ b/internal/proxy/options.go
@@ -48,11 +48,11 @@ import (
 type Options struct {
 	Port int `envconfig:"PORT" default:"4180"`
 
-	ProviderURLString      string `envconfig:"PROVIDER_URL"`
-	ProxyProviderURLString string `envconfig:"PROXY_PROVIDER_URL"`
-	UpstreamConfigsFile    string `envconfig:"UPSTREAM_CONFIGS"`
-	Cluster                string `envconfig:"CLUSTER"`
-	Scheme                 string `envconfig:"SCHEME" default:"https"`
+	ProviderURLString         string `envconfig:"PROVIDER_URL"`
+	ProviderURLInternalString string `envconfig:"PROVIDER_URL_INTERNAL"`
+	UpstreamConfigsFile       string `envconfig:"UPSTREAM_CONFIGS"`
+	Cluster                   string `envconfig:"CLUSTER"`
+	Scheme                    string `envconfig:"SCHEME" default:"https"`
 
 	SkipAuthPreflight bool `envconfig:"SKIP_AUTH_PREFLIGHT"`
 
@@ -132,9 +132,6 @@ func (o *Options) Validate() error {
 	}
 	if o.ProviderURLString == "" {
 		msgs = append(msgs, "missing setting: provider-url")
-	}
-	if o.ProxyProviderURLString == "" {
-		o.ProxyProviderURLString = o.ProviderURLString
 	}
 	if o.UpstreamConfigsFile == "" {
 		msgs = append(msgs, "missing setting: upstream-configs")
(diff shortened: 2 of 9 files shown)


sporkmonger:
Ready for review.

sporkmonger:
I think this could be a table-driven test as well, but would need to edit existing tests rather than creating new ones I think.

internal/proxy/oauthproxy_test.go
```diff
@@ -263,6 +263,129 @@ func TestNewReverseProxyTLSSkipVerify(t *testing.T) {
 	}
 }
 
+func TestNewReverseProxyPreserveHost(t *testing.T) {
```
shrayolacrayon: can we make these table drive tests as well?

internal/proxy/proxy_config.go
```diff
@@ -81,6 +82,7 @@ type OptionsConfig struct {
 	SkipAuthRegex   []string          `yaml:"skip_auth_regex"`
 	AllowedGroups   []string          `yaml:"allowed_groups"`
 	TLSSkipVerify   bool              `yaml:"tls_skip_verify"`
+	PreserveHost    bool              `yaml:"preserve_host"`
```
shrayolacrayon: might be good to add a config unit test as well

sporkmonger:
Hey, sorry about the delay, was heads-down on the Azure AD PR. I’ve got this on my plate next.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

internal/proxy/proxy_config.go
sporkmonger: Done.

internal/proxy/oauthproxy_test.go
sporkmonger: Almost done w/ these and so far confirmed that replacing a bunch of test functions w/ two bigger table-driven test functions hasn't altered code coverage.

sporkmonger: Done.

sporkmonger:
This should be ready to review again.

bertjwregeer:
`X-Forwarded-Host` should be specified multiple times, preferably each time a new proxy forwards it and had received the request using an alternate "Host:" line from an upstream.

Having two is not a bug.

HTTP header folding will change:

```
X-Forwarded-Host: example.com
X-Forwarded-Host: whatever.internal.example.com
```

into:

```
X-Forwarded-Host: example.com, whatever.internal.example.com
```

It is up to client libraries to parse the `X-Forwarded-Host` and use the first value: `example.com` (the proxy closest to the original client), unless it only trusts 1 proxy, in which case it should use `whatever.internal.example.com` instead.

Yes, why do you have to count the proxies you trust, well because the following:

```
X-Forwarded-Host: attacker.com, example.com, whatever.internal.example.com, lb.whatever.example.com
```

And the application is configured to only support 3 proxies, then `attacker.com` should be dropped as it is not trusted, and is likely from an attacker sending an X-Forwarded-Host directly to your first proxy, which is dutifully passing it on as it doesn't strip HTTP headers (Amazon ELB for instance doesn't strip those headers, it also doesn't add an X-Forwarded-Host, but instead passes the Host: header through untouched...).

Please don't remove the multiple `X-Forwarded-Host` headers, unless there is some knob to turn that on/off.

sporkmonger:
I'd advocate for using https://tools.ietf.org/html/rfc7239, which is actually specified, unlike `X-Forwarded-Host`, but sadly it's not well supported yet. But yeah, this probably shouldn't have been in this PR in the first place. Dropped it.

internal/proxy/oauthproxy.go
```diff
@@ -212,9 +214,13 @@ func NewRewriteReverseProxy(route *RewriteRoute, config *UpstreamConfig) *httput
 		}
 		director := httputil.NewSingleHostReverseProxy(target).Director
 
-		req.Header.Add("X-Forwarded-Host", req.Host)
+		if req.Header.Get("X-Forwarded-Host") == "" {
```
bertjwregeer: Here is one more instance where a new X-Forwarded-Host is not added if there is already one in the request.

internal/proxy/oauthproxy_test.go
```diff
@@ -427,7 +427,7 @@ func TestRoundTrip(t *testing.T) {
 	}{
 		{
 			name: "no error",
-			url:  "https://www.example.com/",
+			url:  "http://www.example.com/",
```
sporkmonger: Looks like it's already been fixed actually.

jphines:
@bertjwregeer Thanks for the review! I was going to leave a similar comment regarding the `X-Forwarded-Host` and that I wasn't super comfortable being with it potentially being overridden by the client.

@sporkmonger Going to run some integration tests and see if I can get this merged this afternoon.

bertjwregeer:
@jphines I'm a maintainer for a pure python web server, I've been knee deep in `X-Forwarded-*` stuff for supporting proxies and munging the WSGI environment to match, so I did a drive by review :-)

Issue #40, opened 2018-08-25
Simplify secret validation code


mccutchen:
In #37, I covered some of the difficulties I encountered in generating a valid secret.  A lot of the complexity arises from the fact that sso accepts either base64-encoded bytes or raw bytes, and attempt to guess which it has been given.

Unfortunately, that guess is extremely unreliable, because
 - It will reject legitimate raw secret bytes that _happen_ to decode as base64
 - The base64 decoder used will not decode the output of the `base64` CLI tool a user might naturally reach for to safely encode a set of randomly generated bytes

Here are some legitimate secret values that will be rejected:

```bash
# 32 ASCII bytes generated by 1Password
H3LWy8KF86hm2UcE3XAYvb2ub2G33YiV

$ openssl rand 32 -base64
AquwB7PgQRVaAFKwEXDefRhfegtqfLXVP4/LbDSHkpY=

$ openssl rand 32 | base64
4QkZXXwT27g4tYbdO50gA0UHo5S8OElg0Vs0m+E5w1c=
```

I can't remember why there's so much flexibility built into the secret validation code, but it feels too complex:
https://github.com/buzzfeed/sso/blob/9f61688d4cd0c6252fa347386c64b90eff4cc4fd/internal/auth/options.go#L175-L194

And it seems like this code should be stricter about what it will accept, and rely on the application  to provide secret bytes in some specific form:
https://github.com/buzzfeed/sso/blob/9f61688d4cd0c6252fa347386c64b90eff4cc4fd/internal/pkg/sessions/cookie_store.go#L50-L57

mreiferson:
Looks like this landed in #56, closing!

PR #56, opened 2018-09-18
simplify secret validation

shrayolacrayon:
## Problem
This addresses https://github.com/buzzfeed/sso/issues/40. 

## Solution
This makes validation of cookie secrets more strict, allowing only for a base64 encoded 32-byte secret. 

Cookie secrets should be able to be created using 
`openssl rand 32 -base64` 

cc @buzzfeed/sso-maintainers

### DIFF
--- a/internal/auth/options.go
+++ b/internal/auth/options.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"crypto"
+	"encoding/base64"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -12,7 +13,6 @@ import (
 	"github.com/buzzfeed/sso/internal/auth/providers"
 	"github.com/buzzfeed/sso/internal/pkg/groups"
 	log "github.com/buzzfeed/sso/internal/pkg/logging"
-	"github.com/buzzfeed/sso/internal/pkg/sessions"
 )
 
 // Options are config options that can be set by environment variables
@@ -107,6 +107,7 @@ type Options struct {
 	// internal values that are set after config validation
 	redirectURL         *url.URL
 	provider            providers.Provider
+	decodedCookieSecret []byte
 	GroupsCacheStopFunc func()
 }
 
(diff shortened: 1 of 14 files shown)


internal/proxy/options.go
```diff
@@ -179,25 +179,12 @@ func (o *Options) Validate() error {
...
-			}
-			msgs = append(msgs, fmt.Sprintf("cookie_secret must be 32 or 64 bytes  but is %d bytes.%s",
-				len(secretBytes(o.CookieSecret)), suffix))
-		}
+	decodedCookieSecret, err := base64.StdEncoding.DecodeString(o.CookieSecret)
+	if err != nil {
+		msgs = append(msgs, "invalid cookie secret")
+	}
+	if len(decodedCookieSecret) != 32 {
+		msgs = append(msgs, fmt.Sprintf("invalid cookie secret length, must be 32 bytes but was %d bytes", len(decodedCookieSecret)))
```
mccutchen: All of the comments from the proxy side above apply here, too, w/r/t to error messages and allowing 32 or 64 byte secrets.

internal/auth/options.go
```diff
@@ -171,25 +171,12 @@ func (o *Options) Validate() error {
...
-	if validCookieSecretSize == false {
-		var suffix string
-		if decoded {
-			suffix = fmt.Sprintf(" note: cookie secret was base64 decoded from %q", o.CookieSecret)
-		}
-		msgs = append(msgs, fmt.Sprintf(
-			"cookie_secret must be 32 or 64 bytes "+
-				"to create an AES cipher but is %d bytes.%s",
-			len(sessions.SecretBytes(o.CookieSecret)), suffix))
+	if len(decodedCookieSecret) != 32 {
```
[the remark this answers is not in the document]
shrayolacrayon: No, there's not really a reason not to, can add that back.

internal/pkg/sessions/cookie_store.go
```diff
@@ -68,10 +68,10 @@ func addPadding(secret string) string {
 }
 
 // CreateMiscreantCookieCipher creates a new miscreant cipher with the cookie secret
-func CreateMiscreantCookieCipher(cookieSecret string) func(s *CookieStore) error {
+func CreateMiscreantCookieCipher(cookieSecret []byte) func(s *CookieStore) error {
 	return func(s *CookieStore) error {
-		s.CookieSecret = cookieSecret
-		cipher, err := aead.NewMiscreantCipher(SecretBytes(s.CookieSecret))
+		s.CookieSecret = string(cookieSecret)
```
mccutchen: Is this `CookieSecret` field on `CookieStore` actually needed/used?  From a quick grep through the codebase, it doesn't appear to be used.  Can we drop it?

quickstart/docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
 
       # XXX: These secrets are for demonstration purposes only! Use
       #
-      #     openssl rand -base64 32 | head -c 32 | base64
+      #     openssl rand -base64 32
       #
       # to generate your own.
       - AUTH_CODE_SECRET=SVM0NEFMUUlaZGxyaFVhOGxsQ0wvOFYyZTh2S2Fha1U=
```
jphines: Do these still work in light of our changes above?

internal/auth/options.go
```diff
@@ -171,27 +172,25 @@ func (o *Options) Validate() error {
...
-		if decoded {
-			suffix = fmt.Sprintf(" note: cookie secret was base64 decoded from %q", o.CookieSecret)
-		}
-		msgs = append(msgs, fmt.Sprintf(
-			"cookie_secret must be 32 or 64 bytes "+
-				"to create an AES cipher but is %d bytes.%s",
-			len(sessions.SecretBytes(o.CookieSecret)), suffix))
+
+	if !validCookieSecretLength {
+
```
jphines: Unnecessary newline

PR #57, opened 2018-09-18
proxy: manage session failure with restart of oauth flow

jphines:
## Problem

It is occasionally necessary to modify the session encoding, either as improvements are made or we want to rotate secrets.

When this occurs, since each upstream session is invalidated, it creates an unpleasant user experience where a user gets a `500` on each initial visit to every upstream.

This poor ux adds unnecessary resistance to updating or modifying the session encoding. While schema changes and/or secret rotations should be taken lightly, we should be confident to make those changes with as minimal impact to UX as possible.

## Solution

Instead of rendering a `500` error in loading an invalid session, we should trigger the start of a new OAuth flow. Our implicit login mechanism should mask this error for most applications and users should be seamlessly re-logged in.

## Notes

This will not work for all applications, particularly applications that do not/cannot follow redirects to finish the OAuth flow

### DIFF
--- a/internal/pkg/sessions/cookie_store.go
+++ b/internal/pkg/sessions/cookie_store.go
@@ -2,15 +2,19 @@ package sessions
 
 import (
 	"encoding/base64"
+	"errors"
 	"fmt"
-	"log"
 	"net"
 	"net/http"
 	"strings"
 	"time"
 
 	"github.com/buzzfeed/sso/internal/pkg/aead"
+	log "github.com/buzzfeed/sso/internal/pkg/logging"
 )
+
+// ErrInvalidSession is an error for invalid sessions.
+var ErrInvalidSession = errors.New("invalid session")
 
 // CSRFStore has the functions for setting, getting, and clearing the CSRF cookie
 type CSRFStore interface {
@@ -102,13 +106,14 @@ func NewCookieStore(cookieName string, optFuncs ...func(*CookieStore) error) (*C
 }
 
 func (s *CookieStore) makeCookie(req *http.Request, name string, value string, expiration time.Duration, now time.Time) *http.Cookie {
+	logger := log.NewLogEntry()
 	domain := req.Host
 	if h, _, err := net.SplitHostPort(domain); err == nil {
 		domain = h
 	}
 	if s.CookieDomain != "" {
 		if !strings.HasSuffix(domain, s.CookieDomain) {
-			log.Printf("Warning: Using configured cookie domain. request_host=%s cookie_domain=%s", domain, s.CookieDomain)
+			logger.WithRequestHost(domain).WithCookieDomain(s.CookieDomain).Warn("Warning: Using configured cookie domain.")
 		}
 		domain = s.CookieDomain
 	}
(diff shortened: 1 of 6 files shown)


jphines:
This PR should follow https://github.com/buzzfeed/sso/pull/54 has it is dependent on those changes. 

I've cherry-picked it into this branch to work on these changes until we have the opportunity to merge 54 and I can rebase my changes.

shrayolacrayon:
This looks great! Could we add a test for the going to OAuthStart?

jphines:
@shrayolacrayon I added a test for our authenticate behavior to ensure when our cipher fails to decrypt the session we bubble up the new `InvalidSession` to callers.

It doesn't look like we have great tests around the expected user flows for various edge cases i.e. of the `Proxy` function in particular.

I'd be happy to work on tests for those various flows in a follow-up PR.

internal/pkg/sessions/cookie_store.go
```diff
@@ -167,7 +171,7 @@ func (s *CookieStore) LoadSession(req *http.Request) (*SessionState, error) {
 	}
 	session, err := UnmarshalSession(c.Value, s.CookieCipher)
 	if err != nil {
-		return nil, err
```
jphines: Let's add some logging for this error and/or create some metrics for it.

I don't want to just swallow this error as it may be very useful for debugging context, despite not being useful further up the stack.

internal/pkg/logging/logging.go
```diff
@@ -146,7 +146,7 @@ func (l *LogEntry) WithEndpoint(endpoint string) *LogEntry {
 
 // WithError appends an `error` tag to a LogEntry. Useful for annotating non-Error log
 // entries (e.g. Fatal messages) with an `error` object.
-func (l *LogEntry) WithError(err error) *LogEntry {
+func (l *LogEntry) WithError(err interface{}) *LogEntry {
```
jphines: This type safety here seemed good, why are removing it?

internal/pkg/logging/logging.go
```diff
@@ -50,7 +50,7 @@ func (l *LogEntry) Warn(args ...interface{}) {
 
 // Error wraps the logrus Error function
 func (l *LogEntry) Error(err interface{}, args ...interface{}) {
```
jphines: Maybe we should make this an `error` instead of removing the type safety from `WithError`?

internal/pkg/sessions/cookie_store.go
```diff
@@ -160,14 +165,16 @@ func (s *CookieStore) setSessionCookie(rw http.ResponseWriter, req *http.Request
...
+	logger := log.NewLogEntry()
 	c, err := req.Cookie(s.Name)
 	if err != nil {
 		// always http.ErrNoCookie
 		return nil, err
 	}
 	session, err := UnmarshalSession(c.Value, s.CookieCipher)
 	if err != nil {
-		return nil, err
+		logger.WithRequestHost(req.Host).WithError(err)
```
jphines: I think we need to add a logging action here. IIRC, this doesn't actually log anything but returns a new logger with these default fields, so this might become:

```golang 
logger.WithRequestHost(req.Host).Error(err, "unable to unmarshal session")
```

internal/pkg/logging/logging.go
jphines: Or is this really meant to be a `string`?

shrayolacrayon: We have a lot of places where we're passing in a string with `logger.Error(...)`, I can change those if we want to keep the type safety though.

shrayolacrayon: I'm gonna do this in a separate followup PR

jphines:
@shrayolacrayon let's rebase + squash + merge at your leisure.

PR #58, opened 2018-09-18
circle: fix issues with circle script checking for owner

loganmeetsworld:
Fixes issues introduced in #53


sporkmonger:
That seems to have fixed it, thanks!

PR #63, opened 2018-09-19
Static assets now built into the binary

### LINKED ISSUE #7: build static assets into go binary
**Is your feature request related to a problem? Please describe.**
Currently, static assets are not built into the binary.

**Describe the solution you'd like**
We can build the static assets into the go binary to make it easier to pull and use the Docker image without needing to mount the static directory. 
We can use [go-bindata](https://github.com/jteeuwen/go-bindata) to compile static images into our go binary, similar to what is done in [go-httpbin](https://github.com/mccutchen/go-httpbin)

sporkmonger:
## Problem

Fixes #7, simplifies deployment.

## Solution

Bake static assets into a `.go` file that functions as a file server.

## Notes

Opted to use https://github.com/rakyll/statik over https://github.com/jteeuwen/go-bindata since it's actively maintained.


internal/auth/authenticator.go
```diff
@@ -180,8 +183,12 @@ func (p *Authenticator) newMux() http.Handler {
 	serviceMux.HandleFunc("/validate", p.withMethods(p.validateClientID(p.validateClientSecret(p.ValidateToken)), "GET"))
 	serviceMux.HandleFunc("/redeem", p.withMethods(p.validateClientID(p.validateClientSecret(p.Redeem)), "POST"))
 	serviceMux.HandleFunc("/refresh", p.withMethods(p.validateClientID(p.validateClientSecret(p.Refresh)), "POST"))
-	// serve static files but prevent directory listings, see static_files.go
-	serviceMux.Handle("/static/", http.StripPrefix("/static/", http.FileServer(noDirectoryFS{http.Dir("./static")})))
+	fsHandler, err := loadFSHandler()
+	if err != nil {
+		logger.Fatal(err)
+		os.Exit(1)
```
sporkmonger: Actually, I meant to check whether `.Fatal` calls `os.Exit` or not before committing. Looks like it does, so this line isn't needed.

sporkmonger: That said this is one of those "this should never happen" errors.

sporkmonger:
I opted to check `internal/auth/statik/statik.go` into source control so that you wouldn't need to call `go generate` immediately after a clone to get the project to compile and because I suspect it won't change often. But I could also easily see the argument in the other direction since it's a generated file.

sporkmonger:
So it seems like the `gpm install` in the earlier build step maybe isn't doing what I'm expecting it to?

sporkmonger:
OK, closer, but maybe an issue with wrong version of Go?

sporkmonger:
Failures that only happen in CI/CD are always fun.

sporkmonger:
I'm going to squash the fighting-with-CircleCI commits.

sporkmonger:
Ready for review.

shrayolacrayon:
Thanks for opening @sporkmonger! We'll be working on reviewing and getting this merged as soon as we can.

shrayolacrayon:
One more request - we should be able to also get rid of copying over the static directory in the [Dockerfile](https://github.com/buzzfeed/sso/blob/6bc3c4473d78e94c00775f2297c3a540c7daf0d5/Dockerfile#L31) when building the docker image. 

Since we are still working on figuring out a good way to build and push docker images for forked community contributions, I've pushed a copy of your pr branch on buzzfeed/sso (https://github.com/buzzfeed/sso/tree/pr/63) to validate on our internal infrastructure so that we can get this merged!

sporkmonger:
Good point, I’ll take care of it.

shrayolacrayon:
Thanks for making those changes @sporkmonger, this looks good to me, can you please squash the commits and I'll approve and merge!

sporkmonger:
All squashed!

PR #71, opened 2018-09-25
proxy: logging handling/logging of oauth start on xml requests

jphines:
## Problem

XHR error handling was creating obscure corner cases due to some bugs in the request flow.

## Solution

Improve logging to better note handling of these requests as well as fix a bug in the request flow.


shrayolacrayon:
This looks great! Can we add a test?

jphines:
@shrayolacrayon added a test for the `nil` cookie case

shrayolacrayon:
nice this lgtm!

PR #76, opened 2018-09-27
Refactor provider options

### LINKED ISSUE #6: sso-auth: decouple setting provider with options validation
## What 
Move out initializing providers from validating Options into to a separate function that can be taken in as an option function. 
## Why 
Currently we set the provider as the google provider during [validation](https://github.com/buzzfeed/sso/blob/master/internal/auth/options.go#L176) of the `Options` struct which makes it difficult to add a new provider and to add new features to existing providers. Decoupling the logic from validating Options will make it simpler to support adding new providers.

sporkmonger:
Introduces a simple `switch` to initialize the provider to allow for new providers to be subsequently added.

I did consider removing the `TestGoogleGroupInvalidFile` function and including it in `TestNewOptions`, since it no longer requires a `panic` recovery, but that seemed like maybe the wrong solution once multiple providers are introduced.

Fixes #6.

### DIFF
--- a/internal/auth/authenticator_test.go
+++ b/internal/auth/authenticator_test.go
@@ -1541,7 +1541,9 @@ func TestOAuthStart(t *testing.T) {
 			opts := testOpts("abced", "testtest")
 			opts.RedirectURL = "https://example.com/oauth2/callback"
 			opts.Validate()
-			proxy, _ := NewAuthenticator(opts, func(p *Authenticator) error {
+			u, _ := url.Parse("http://example.com")
+			provider := providers.NewTestProvider(u)
+			proxy, _ := NewAuthenticator(opts, setTestProvider(provider), func(p *Authenticator) error {
 				p.Validator = func(string) bool { return true }
 				return nil
 			}, setMockCSRFStore(&sessions.MockCSRFStore{}))
@@ -1631,4 +1633,46 @@ func TestHostHeader(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestGoogleProviderApiSettings(t *testing.T) {
+	opts := testOpts("abced", "testtest")
+	opts.Provider = "google"
+	opts.Validate()
+	proxy, _ := NewAuthenticator(opts, AssignProvider(opts), func(p *Authenticator) error {
+		p.Validator = func(string) bool { return true }
+		return nil
+	})
+	p := proxy.provider.Data()
+	testutil.Equal(t, "https://accounts.google.com/o/oauth2/auth?access_type=offline",
+		p.SignInURL.String())
+	testutil.Equal(t, "https://www.googleapis.com/oauth2/v3/token",
+		p.RedeemURL.String())
+	testutil.Equal(t, "", p.ProfileURL.String())
+	testutil.Equal(t, "profile email", p.Scope)
+}
+
+func TestGoogleGroupInvalidFile(t *testing.T) {
+	opts := testOpts("abced", "testtest")
+	opts.Provider = "google"
+	opts.GoogleAdminEmail = "admin@example.com"
+	opts.GoogleServiceAccountJSON = "file_doesnt_exist.json"
+	opts.Validate()
+	_, err := NewAuthenticator(opts, AssignProvider(opts), func(p *Authenticator) error {
+		p.Validator = func(string) bool { return true }
+		return nil
+	})
+	testutil.NotEqual(t, nil, err)
+	testutil.Equal(t, "invalid Google credentials file: file_doesnt_exist.json", err.Error())
+}
+
+func TestUnimplementedProvider(t *testing.T) {
+	opts := testOpts("abced", "testtest")
+	opts.Validate()
+	_, err := NewAuthenticator(opts, AssignProvider(opts), func(p *Authenticator) error {
+		p.Validator = func(string) bool { return true }
+		return nil
+	})
+	testutil.NotEqual(t, nil, err)
+	testutil.Equal(t, "unimplemented provider: \"\"", err.Error())
 }
--- a/internal/auth/options.go
+++ b/internal/auth/options.go
@@ -106,7 +106,6 @@ type Options struct {
 
 	// internal values that are set after config validation
 	redirectURL         *url.URL
-	provider            providers.Provider
 	decodedCookieSecret []byte
 	GroupsCacheStopFunc func()
 }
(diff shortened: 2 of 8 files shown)


sporkmonger:
Ready for review.

shrayolacrayon:
Thanks for starting on this @sporkmonger, I think I had a deeper refactor in mind for this issue, which is to remove all provider logic from options validation.  

I would approach this starting off with removing the `provider` field from `Options` completely and initializing the the Provider in `NewAuthenticator`  and using a functional option to assign the provider to the Authenticator, as that is something we are trying to move towards as well. (see https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)

sporkmonger:
I hadn't ever seen that article before, thanks for that. That's super-cool.

sporkmonger:
@shrayolacrayon I started on making the changes, but given the order in which options are processed, I'm not sure how this would work in a `NewOptions` context.

https://github.com/buzzfeed/sso/blob/master/cmd/sso-auth/main.go#L21-L33

I was thinking, e.g.

```go
opts, err := auth.NewOptions(options.GoogleProvider("admin@domain.com", "credentials.json"))
```

But the provider's not known at the point where you call `NewOptions`?

shrayolacrayon:
Hmm I had in mind doing something like this, where we validate the non-provider options in `options.Validate` and validate and create the provider in a separate step when we're initializing the Authenticator. 

```
provider.go

func AssignProvider(opts *Option)func(*Authenticator)error{
return func(a *Authenticator)error{
  // initialize the provider and validate options in NewProvider
  provider, err := NewProvider(options)
  if err != nil {
    return err
  }
  a.Provider = provider
}}

main.go 

func main(){
...
// initialize all options, validate all options that are not provider specific 
options := NewOptions()
err := envconfig.options.Process()
err := options.Validate() // This does NOT validate any of the provider data 


// AssignProvider
authenticator, err := NewAuthenticator(options, authenticator.AssignProvider, ....)
}

```

sporkmonger:
`authenticator.AssignProvider` seems like it would be an always-required parameter in this scenario? That seems odd to me. It might be beneficial for testing purposes, but otherwise it strikes me as harder to understand and reason about than a simple helper function. I really like this pattern, but it seems mostly applicable when you have a larger number of optional bits of configuration you're applying, particularly where they may have their own context-sensitive parameters. In this case we're always choosing a provider (except for maybe in some test scenarios?). I'm not sure if there are other config functions to pass in to `NewAuthenticator` besides `authenticator.AssignProvider`, but it doesn't appear like there would be in that design?

The pattern made more sense to me as a way to minimize the complexity of the provider `switch` logic, since you could do something like:

```go
switch opts.Provider {
case "azure":
  providerConfig = authenticator.AzureProvider(opts.AzureTenantID)
case "google":
  providerConfig = authenticator.GoogleProvider(opts.GoogleAdminEmail, opts.GoogleServiceAccountJSON)
}

authenticator, err := NewAuthenticator(options, providerConfig, ....)
```

Writing it as `authenticator.AssignProvider` doesn't feel quite right to me.

shrayolacrayon:
The `switch` logic definitely makes sense to me, and would make sense to be encapsulated in a `NewProvider` function ie:

` internal/auth/options.go`
```go

var (
GoogleProvider = "google"
AzureProvider = "azure"
)

func NewProvider(opts) (providers.Provider, error) {
switch opts.Provider {
  case GoogleProvider: 
    return providers.NewGoogleProvider(opts)
  case AzureProvider:
    return providers.NewAzureProvider(opts)
  default:
    return nil, fmt.Errorf("unsupported provider in options: %s", opts.Provider)
}

}
```

I think assigning the provider in a functional option could be useful for some test cases and is a pattern that we have been using for mocking in tests, but I'm not attached to the idea of using it for the providers. FWIW there are already functional options being passed into [`NewAuthenticator`](https://github.com/buzzfeed/sso/blob/master/cmd/sso-auth/main.go#L40) for the cookie store and for the statsd client (the `CookieStore` logic has not yet been migrated over to `sso-proxy` because it was a more recent change before the open-sourcing) so it would not be a new pattern to use.

sporkmonger:
OK, cool, I think that gives me enough to work with.

sporkmonger:
It might also be a good pattern for configuring statsd vs prometheus & opentracing too, FWIW.

sporkmonger:
OK, I removed the `provider` variable from `Options` entirely, it now only lives in `Authenticator`, so a couple tests shifted over from `options_test.go` to `authenticator_test.go`.

internal/auth/options.go
```diff
@@ -227,42 +227,74 @@ func (o *Options) Validate() error {
...
+		return nil, err
+	}
 
-	if o.GoogleServiceAccountJSON != "" {
-		_, err := os.Open(o.GoogleServiceAccountJSON)
-		if err != nil {
-			msgs = append(msgs, "invalid Google credentials file: "+o.GoogleServiceAccountJSON)
+	var singleFlightProvider providers.Provider
+	switch o.Provider {
+	default: // Google
```
shrayolacrayon: can we make the default return an error here rather than the google provider and add a separate switch case for google? We can for now set the provider in `NewOptions()` to be google.

sporkmonger: Sure. I actually had it that way initially but it broke half the test suite, but it's probably not a bad idea to make the tests more explicit.

internal/auth/authenticator_test.go
```diff
@@ -1631,3 +1633,32 @@ func TestHostHeader(t *testing.T) {
...
+	testutil.Equal(t, "https://accounts.google.com/o/oauth2/auth?access_type=offline",
+		p.SignInURL.String())
+	testutil.Equal(t, "https://www.googleapis.com/oauth2/v3/token",
+		p.RedeemURL.String())
+	testutil.Equal(t, "", p.ProfileURL.String())
+	testutil.Equal(t, "profile email", p.Scope)
+}
+
+func TestGoogleGroupInvalidFile(t *testing.T) {
+	opts := testOpts("abced", "testtest")
```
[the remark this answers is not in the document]
sporkmonger: I'd lean towards error rather than panic.

internal/auth/providers/providers.go
```diff
@@ -24,6 +24,23 @@ var (
 	ErrServiceUnavailable = errors.New("SERVICE_UNAVAILABLE")
 )
 
+const (
+	// AzureProviderName identifies the Azure AD provider
+	AzureProviderName = "azure"
```
shrayolacrayon: I'm not sure if we need all of these constants yet, might be good to just keep Google for now and add them as the providers have been implemented as to not confused anyone that they have been.

sporkmonger: Sure, that seems fine. I'm planning to do these next-ish, but they can go into another PR instead, that's fine.

shrayolacrayon:
This looks good to me! Can you squash and we can merge?

sporkmonger:
Should be all set!

shrayolacrayon:
Thanks for doing this @sporkmonger!

sporkmonger:
Sure thing!

PR #77, opened 2018-10-01
pkg/aead: add explicit data race and locking to resolve

jphines:
This adds an explicit concurrency test case for aead/miscreant to test for potential data races in the underlying cipher implementation.

### DIFF
--- a/internal/pkg/aead/aead.go
+++ b/internal/pkg/aead/aead.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"sync"
 
 	miscreant "github.com/miscreant/miscreant-go"
 )
@@ -26,6 +27,8 @@ type Cipher interface {
 // For a description of the methodology, see https://en.wikipedia.org/wiki/Authenticated_encryption
 type MiscreantCipher struct {
 	aead cipher.AEAD
+
+	mux sync.Mutex
 }
 
 // NewMiscreantCipher returns a new AES Cipher for encrypting values
@@ -46,6 +49,9 @@ func GenerateKey() []byte {
 
 // Encrypt a value using AES-CMAC-SIV
 func (c *MiscreantCipher) Encrypt(plaintext []byte) (joined []byte, err error) {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+
 	defer func() {
 		if r := recover(); r != nil {
 			err = fmt.Errorf("miscreant error encrypting bytes: %v", r)
@@ -61,6 +67,9 @@ func (c *MiscreantCipher) Encrypt(plaintext []byte) (joined []byte, err error) {
 
 // Decrypt a value using AES-CMAC-SIV
 func (c *MiscreantCipher) Decrypt(joined []byte) ([]byte, error) {
+	c.mux.Lock()
+	defer c.mux.Unlock()
+
 	if len(joined) <= miscreantNonceSize {
 		return nil, fmt.Errorf("invalid input size: %d", len(joined))
 	}
--- a/internal/pkg/aead/aead_test.go
+++ b/internal/pkg/aead/aead_test.go
@@ -1,7 +1,11 @@
 package aead
 
 import (
+	"crypto/rand"
+	"crypto/sha1"
+	"fmt"
 	"reflect"
+	"sync"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


internal/pkg/aead/aead.go
```diff
@@ -61,6 +67,11 @@ func (c *MiscreantCipher) Encrypt(plaintext []byte) (joined []byte, err error) {
 
 // Decrypt a value using AES-CMAC-SIV
 func (c *MiscreantCipher) Decrypt(joined []byte) ([]byte, error) {
+	aead, err := miscreant.NewAEAD(algorithmType, c.secret, miscreantNonceSize)
```
katzdm: It's worth pointing out that this seems to be a somewhat nontrivial initializer, which performs some block cipher encryption to derive its initial state, on each call.
https://github.com/miscreant/miscreant-go/blob/master/cmac/cmac.go#L45

My assumption is that `sso_proxy` is not particularly compute-bounded, but it could be interesting to throw this up in `stage`, and using Datadog to verify that we don't see an undue latency cost from these initializations.

jphines:
@katzdm Do you mean something like https://circleci.com/gh/buzzfeed/sso/463 which is shows the data race?

internal/pkg/aead/aead.go
```diff
@@ -25,17 +25,18 @@ type Cipher interface {
 // Using an AEAD is a cipher providing authenticated encryption with associated data.
 // For a description of the methodology, see https://en.wikipedia.org/wiki/Authenticated_encryption
 type MiscreantCipher struct {
-	aead cipher.AEAD
+	secret []byte
 }
 
 // NewMiscreantCipher returns a new AES Cipher for encrypting values
 func NewMiscreantCipher(secret []byte) (*MiscreantCipher, error) {
```
katzdm: This is a somewhat minor point, but it feels surprising for this type to be named `MiscreantCipher`, if the object in question (the Miscreant cipher) is only instantiated at `Encrypt`/`Decrypt`-time. Maybe it would be better if we keep the `cipher.AEAD` object owned by the `MiscreantCipher` class, and instead initialize a new `MiscreantCipher` during `Marshal`/`Unmarsal`. Might make it easier to reason about the types, in the future.

What do you think?

jphines: I'm okay with changing the type name, but I think whatever we want to expose, we want to be `go-routine` safe. I would argue that this should include making both `Encrypt` and `Decrypt` go-routine safe which currently necessitates the instantiation of new objects.

jphines: I think _just_ creating new objects in the `Unmarshal` and `Marshal` such that those are the only two methods that would be go-routine safe isn't sufficient for what we want to expose from this API.

shrayolacrayon: FWIW the `MiscreantCipher` is an implementation of the `Cipher` interface (which all might be poorly named) and has `Unmarshal` and `Marshal` as well as `Encrypt` and `Decrypt` functions. This probably could be structured differently but  we'd just need to reflect those changes in the `MockCipher`

internal/pkg/aead/aead.go
jphines: I suppose instead of instantiating new objects we could just lock around a single cipher?

I wonder which would be more performant.

internal/pkg/aead/aead_test.go
```diff
@@ -89,3 +93,72 @@ func TestMarshalAndUnmarshalStruct(t *testing.T) {
 		t.Fatalf("expected structs to be equal")
 	}
 }
+
+func TestCipherDataRace(t *testing.T) {
```
katzdm: Consider annotating with a comment pointing to a GitHub issue describing the problem. Six months from now, we'll all forget why we added this test :-)

jphines: Added

internal/pkg/aead/aead_test.go
```diff
@@ -89,3 +93,79 @@ func TestMarshalAndUnmarshalStruct(t *testing.T) {
 		t.Fatalf("expected structs to be equal")
 	}
 }
+
+// TestCipherDataRace exercises a simple concurrency test for the MicscreantCipher.
+// In https://github.com/buzzfeed/sso/pull/75 we investigated why, on random occasion,
+// unmarshalling session states would fail, triggering users to get kicked out of
+// authenticated states. We narrowed our suspicious to a data race that we uncovered
```
katzdm: nit: s/suspicious/suspicions

Also, I don't think we should describe the miscreant library as "having a data race"; the library just doesn't make any attempt at thread-safety. It might be more accurate to phrase the situation as something like: "We eventually uncovered that `sso` implicitly assumed the underlying miscreant APIs to be thread-safe; in truth, they are only thread-compatible and require external synchronization."

shrayolacrayon:
this LGTM, lets squash + rebase and merge?

jphines:
@shrayolacrayon squashed into two commits + rebased.

Issue #26, opened 2018-08-23
add a new parameter which overrides the `PROVIDER_URL` for the sso-proxy->sso-auth connection for split dns environments


danbf:
add a new parameter which overrides the `PROVIDER_URL` for the sso-proxy->sso-auth connection for split dns environments

currently the `PROVIDER_URL` is both used to generate the sign_in url for the client `http://sso-auth.localtest.me/sign_in?` and by the sso-proxy for the redeem transaction `http://sso-auth.localtest.me/redeem`.  In a split DNS environment they point to different IP's as is the case with the current quickstart.

currently `sso-auth.localtest.me` resolves to different IP's for the client and the sso-proxy process:

for `PROVIDER_URL=sso-auth.localtest.me`
Client Resolution: sso-auth.localtest.me=127.0.0.1
sso-proxy Resolution: sso-auth.localtest.me=172.20.0.1 which forced by https://github.com/buzzfeed/sso/blob/master/quickstart/docker-compose.yml#L64-L65 and https://github.com/buzzfeed/sso/blob/master/quickstart/docker-compose.yml#L137-L143

i propose an optional parameter `PROXY_PROVIDER_URL` which when present is used by the sso-proxy to generate the redeem transaction url.  when not present the current behavior of `PROVIDER_URL` being used.  ideally this would result in:

for `PROVIDER_URL=sso-auth.localtest.me`
for `PROXY_PROVIDER_URL=host.docker.internal`
Client Resolution: sso-auth.localtest.me=127.0.0.1
sso-proxy Resolution: host.docker.internal=172.20.0.1
with out the need for any docker-compose `extra_hosts` or `networks` stanza's

tyrken:
FYI just hit this in an AWS K8s deploy, and glad I saw you'd already started before i try my own PR.  I'm wanting cluster-internal traffic between sso-proxy and sso-auth to stay inside the cluster, not need to go out & back in via the locked down external ingress.

danbf:
resolved in https://github.com/buzzfeed/sso/pull/88

PR #88, opened 2018-10-08
sso-proxy: adding an optional PROXY_PROVIDER_URL for split dns deployments

danbf:
sso-proxy: adding an optional PROXY_PROVIDER_URL for split dns deployments

addressing: https://github.com/buzzfeed/sso/issues/26

### DIFF
--- a/internal/proxy/options.go
+++ b/internal/proxy/options.go
@@ -19,6 +19,7 @@ import (
 // Options are configuration options that can be set by Environment Variables
 // Port - int -  port to listen on for HTTP clients
 // ProviderURLString - the URL for the provider in this environment: "https://sso-auth.example.com"
+// ProxyProviderURLString - the internal URL for the provider in this environment: "https://sso-auth-int.example.com"
 // UpstreamConfigsFile - the path to upstream configs file
 // Cluster - the cluster in which this is running, used for upstream configs
 // Scheme - the default scheme, used for upstream configs
@@ -46,10 +47,11 @@ import (
 type Options struct {
 	Port int `envconfig:"PORT" default:"4180"`
 
-	ProviderURLString   string `envconfig:"PROVIDER_URL"`
-	UpstreamConfigsFile string `envconfig:"UPSTREAM_CONFIGS"`
-	Cluster             string `envconfig:"CLUSTER"`
-	Scheme              string `envconfig:"SCHEME" default:"https"`
+	ProviderURLString      string `envconfig:"PROVIDER_URL"`
+	ProxyProviderURLString string `envconfig:"PROXY_PROVIDER_URL"`
+	UpstreamConfigsFile    string `envconfig:"UPSTREAM_CONFIGS"`
+	Cluster                string `envconfig:"CLUSTER"`
+	Scheme                 string `envconfig:"SCHEME" default:"https"`
 
 	SkipAuthPreflight bool `envconfig:"SKIP_AUTH_PREFLIGHT"`
 
@@ -124,6 +126,9 @@ func (o *Options) Validate() error {
 	}
 	if o.ProviderURLString == "" {
 		msgs = append(msgs, "missing setting: provider-url")
+	}
+	if o.ProxyProviderURLString == "" {
+		o.ProxyProviderURLString = o.ProviderURLString
 	}
 	if o.UpstreamConfigsFile == "" {
 		msgs = append(msgs, "missing setting: upstream-configs")
@@ -215,10 +220,19 @@ func parseProviderInfo(o *Options) error {
 		return errors.New("provider-url must include scheme and host")
 	}
 
+	proxyProviderURL, err := url.Parse(o.ProxyProviderURLString)
+	if err != nil {
+		return err
+	}
+	if proxyProviderURL.Scheme == "" || proxyProviderURL.Host == "" {
+		return errors.New("proxy provider url must include scheme and host")
+	}
+
 	providerData := &providers.ProviderData{
 		ClientID:           o.ClientID,
 		ClientSecret:       o.ClientSecret,
 		ProviderURL:        providerURL,
+		ProxyProviderURL:   proxyProviderURL,
 		Scope:              o.Scope,
 		SessionLifetimeTTL: o.SessionLifetimeTTL,
 		SessionValidTTL:    o.SessionValidTTL,
--- a/internal/proxy/options_test.go
+++ b/internal/proxy/options_test.go
@@ -71,6 +71,8 @@ func TestDefaultProviderApiSettings(t *testing.T) {
 		p.SignOutURL.String())
 	testutil.Equal(t, "https://www.example.com/redeem",
 		p.RedeemURL.String())
+	testutil.Equal(t, "https://www.example.com/redeem",
+		p.ProxyRedeemURL.String())
 	testutil.Equal(t, "https://www.example.com/validate",
 		p.ValidateURL.String())
 	testutil.Equal(t, "https://www.example.com/profile",
@@ -80,10 +82,12 @@ func TestDefaultProviderApiSettings(t *testing.T) {
 
 func TestProviderURLValidation(t *testing.T) {
 	testCases := []struct {
-		name              string
-		providerURLString string
-		expectedError     string
-		expectedSignInURL string
+		name                           string
+		providerURLString              string
+		proxyProviderURLString         string
+		expectedError                  string
+		expectedProxyProviderURLString string
+		expectedSignInURL              string
 	}{
 		{
 			name:              "http scheme preserved",
@@ -94,6 +98,17 @@ func TestProviderURLValidation(t *testing.T) {
 			name:              "https scheme preserved",
 			providerURLString: "https://provider.example.com",
 			expectedSignInURL: "https://provider.example.com/sign_in",
+		},
+		{
+			name:                           "proxy provider url string based on providerURL",
+			providerURLString:              "https://provider.example.com",
+			expectedProxyProviderURLString: "https://provider.example.com",
+		},
+		{
+			name:                           "proxy provider url string based on proxyProviderURL",
+			providerURLString:              "https://provider.example.com",
+			proxyProviderURLString:         "https://provider-internal.example.com",
+			expectedProxyProviderURLString: "https://provider-internal.example.com",
 		},
 		{
 			name:              "scheme required",
@@ -121,6 +136,7 @@ func TestProviderURLValidation(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			o := testOptions()
 			o.ProviderURLString = tc.providerURLString
+			o.ProxyProviderURLString = tc.proxyProviderURLString
 			err := o.Validate()
 			if tc.expectedError != "" {
 				if err == nil {
@@ -130,10 +146,10 @@ func TestProviderURLValidation(t *testing.T) {
 				}
 			}
 			if tc.expectedSignInURL != "" {
-				p := o.provider.Data()
-				if p.SignInURL.String() != tc.expectedSignInURL {
-					t.Errorf("expected SignInURL = %q, got %q", tc.expectedSignInURL, p.SignInURL.String())
-				}
+				testutil.Equal(t, o.provider.Data().SignInURL.String(), tc.expectedSignInURL)
+			}
+			if tc.expectedProxyProviderURLString != "" {
+				testutil.Equal(t, o.provider.Data().ProxyProviderURL.String(), tc.expectedProxyProviderURLString)
 			}
 		})
 	}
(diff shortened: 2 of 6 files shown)


danbf:
working on this error maybe on the sso_auth side:

```
sso-proxy_1    | 2018/10/08 21:52:16 hardcode.RedeemURL! http://host.docker.internal/redeem
sso-proxy_1    | 2018/10/08 21:52:16 p.RedeemURL! http://sso-auth.localtest.me/redeem
sso-auth_1     | {"action":"redeem","http_status":404,"level":"info","msg":"","proxy_host":"httpbin.sso.localtest.me","remote_address":"172.22.0.1","request_duration":0.078322,"request_method":"POST","request_uri":"/redeem","service":"sso-authenticator","time":"2018-10-08 21:52:16.1089","user":"","user_agent":"sso_proxy/HEAD"}
nginx-proxy_1  | nginx.1    | host.docker.internal 172.22.0.1 - - [08/Oct/2018:21:52:16 +0000] "POST /redeem HTTP/1.1" 404 19 "-" "sso_proxy/HEAD"
sso-proxy_1    | {"error":"got 404 from \"http://sso-auth.localtest.me/redeem\" 404 page not found\n","level":"error","msg":"error redeeming authorization code","remote_address":"172.22.0.1","service":"sso-proxy","time":"2018-10-08 21:52:16.1089"}
sso-proxy_1    | {"http_status":500,"level":"info","msg":"error page","page_message":"Internal Error","page_title":"Internal Error","remote_address":"172.22.0.1","service":"sso-proxy","time":"2018-10-08 21:52:16.1089"}
```

danbf:
ok, so this is working now.  now need to actually pass the values from the config to the function.

danbf:
ok, got values making their way into the functions, just got to fixup tests.

danbf:
@shrayolacrayon running into a problem with these tests: https://github.com/buzzfeed/sso/blob/1b23cb75fa9ddb4ea1343321c34ed072b9f304b5/internal/proxy/providers/sso_test.go#L259-L269

it's not clear to me how they are testing the redeem's and what hostnames/urls it's using to do that.

danbf:
now with some help from @katzdm the tests are working 🎉

danbf:
RFR: @buzzfeed/infra-security

internal/proxy/options.go
```diff
@@ -125,6 +127,9 @@ func (o *Options) Validate() error {
 	if o.ProviderURLString == "" {
 		msgs = append(msgs, "missing setting: provider-url")
 	}
+	if o.ProxyProviderURLString == "" {
```
danbf: not positive this is the best place to do this.  first time in this codebase though.

internal/proxy/providers/sso.go
```diff
@@ -105,10 +106,11 @@ func (p *SSOProvider) Redeem(redirectURL, code string) (*SessionState, error) {
 	params.Add("code", code)
 	params.Add("grant_type", "authorization_code")
 
-	req, err := newRequest("POST", p.RedeemURL.String(), bytes.NewBufferString(params.Encode()))
+	req, err := newRequest("POST", p.ProxyRedeemURL.String(), bytes.NewBufferString(params.Encode()))
 	if err != nil {
 		return nil, err
 	}
+	req.Host = p.RedeemURL.Host
```
shrayolacrayon: Trying to understand the necessity for the ProxyRedeemURL - why are we setting the request host to the RedeemURL's host here?

danbf: the sso-auth is making decisions on serving the content based on the host.  doing this let's the sso-auth know that the traffic is for that http server.  i didn't dig into it on the sso-auth end and figured it would be easier to set this host header here then to require the sso-auth to also have the `PROXY_PROVIDER_URL` var fed to it so it could know addtional virtual server addresses it should be responding to.  

without this host header set the sso-auth throws a 4xx error and the sso-proxy redeem fails.

danbf: I think that's here on the sso-auth end? https://github.com/buzzfeed/sso/blob/master/internal/auth/authenticator.go#L603

internal/proxy/options.go
shrayolacrayon: this seems fine to me, can we add a test for this behavior?

danbf: i stuck in a test for this

internal/proxy/options.go
```diff
@@ -215,10 +220,19 @@ func parseProviderInfo(o *Options) error {
 		return errors.New("provider-url must include scheme and host")
 	}
 
+	proxyproviderURL, err := url.Parse(o.ProxyProviderURLString)
```
benjsto: nitpick here, but should maybe be cased as `proxyProviderURL`

danbf: changed it 👍

internal/proxy/providers/sso_test.go
```diff
@@ -201,11 +210,12 @@ func TestSSOProviderGroups(t *testing.T) {
 
 func TestSSOProviderGetEmailAddress(t *testing.T) {
```
shrayolacrayon: nit, unrelated to the changes but this function name should probably be TestRedeem since it doesn't call `GetEmailAddress` and calls the Redeem function.

internal/proxy/providers/sso_test.go
```diff
@@ -230,6 +240,24 @@ func TestSSOProviderGetEmailAddress(t *testing.T) {
...
+		{
+			Name: "redeem-internal successful",
+			Code: "code1234",
+			RedeemResponse: &redeemResponse{
+				AccessToken:  "a1234",
+				ExpiresIn:    10,
+				RefreshToken: "refresh12345",
+				Email:        "michael.bland@gsa.gov",
+			},
+			ProxyProviderURL: &url.URL{
```
shrayolacrayon: I think what you'd want to do to test this behavior is create a `ProxyProviderResponse` similar to the Redeem Response and Profile Response, as well as a `ProxyRedeemResponse`, and assign the server urls to those provider fields. Then you can test that the endpoints are being hit based on the response.

internal/proxy/providers/sso_test.go
```diff
@@ -241,8 +269,10 @@ func TestSSOProviderGetEmailAddress(t *testing.T) {
 				body, err := json.Marshal(tc.RedeemResponse)
 				testutil.Equal(t, nil, err)
 				p.RedeemURL, redeemServer = newTestServer(http.StatusOK, body)
+				p.ProxyRedeemURL, redeemServer = newTestServer(http.StatusOK, body)
 			} else {
 				p.RedeemURL, redeemServer = newCodeTestServer(400)
+				p.ProxyRedeemURL, redeemServer = newCodeTestServer(400)
```
benjsto: similar to @shrayolacrayon 's comment above, why do we need to create the test server twice here?

internal/proxy/providers/sso_test.go
danbf: changing it

internal/proxy/providers/sso_test.go
```diff
@@ -271,6 +304,100 @@ func TestSSOProviderGetEmailAddress(t *testing.T) {
...
+			}
+			defer profileServer.Close()
+			defer profileServerInternal.Close()
+
+			session, err := p.Redeem("http://redirect/", tc.Code)
+			if tc.RedeemResponseInternal != nil {
+				testutil.Equal(t, nil, err)
+				testutil.NotEqual(t, session, nil)
+				testutil.Equal(t, tc.RedeemResponseInternal.Email, session.Email)
+				testutil.Equal(t, tc.RedeemResponseInternal.AccessToken, session.AccessToken)
```
danbf: i think i have a problem here getting the `internal` server to send back the correct body

danbf: fixed it

internal/proxy/providers/sso_test.go
```diff
@@ -241,8 +269,10 @@ func TestSSOProviderGetEmailAddress(t *testing.T) {
 				body, err := json.Marshal(tc.RedeemResponse)
 				testutil.Equal(t, nil, err)
 				p.RedeemURL, redeemServer = newTestServer(http.StatusOK, body)
```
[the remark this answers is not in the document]
danbf: simplified to clean that up

internal/proxy/options_test.go
```diff
@@ -95,6 +98,17 @@ func TestProviderURLValidation(t *testing.T) {
...
 		},
+		{
+			name:              "redeem string based on providerURL",
+			providerURLString: "https://provider.example.com",
+			expectedRedeemURL: "https://provider.example.com/redeem",
+		},
+		{
+			name:                   "redeem string based on proxyProviderURL",
+			providerURLString:      "https://provider.example.com",
+			proxyProviderURLString: "https://provider-internal.example.com",
```
shrayolacrayon: is this field being used anywhere?

danbf: @shrayolacrayon it's the string that we use to generate the `proxyProviderURL` url 
https://github.com/buzzfeed/sso/blob/dec558663391454c352bf0561a04d604ad5c8372/internal/proxy/options.go#L223-L226 which is used to generate the redeem url: https://github.com/buzzfeed/sso/blob/dec558663391454c352bf0561a04d604ad5c8372/internal/proxy/providers/sso.go#L63

internal/proxy/options_test.go
```diff
@@ -16,7 +16,7 @@ func testOptions() *Options {
 	o.ClientSecret = "xyzzyplugh"
 	o.EmailDomains = []string{"*"}
 	o.ProviderURLString = "https://www.example.com"
-	o.ProxyProviderURLString = "https://internal.example.com"
+	o.ProxyProviderURLString = ""
```
shrayolacrayon: you don't need to set this to be empty

internal/proxy/options_test.go
```diff
@@ -135,6 +152,12 @@ func TestProviderURLValidation(t *testing.T) {
 					t.Errorf("expected SignInURL = %q, got %q", tc.expectedSignInURL, p.SignInURL.String())
 				}
 			}
+			if tc.expectedProxyProviderURLString != "" {
```
shrayolacrayon: this is equivalent to `testutil.Equal(t, o.provider.Data().ProxyProviderURL.String(), tc.expectedProxyProviderURLString)`

danbf: switched to use `testutil.Equal`

internal/proxy/options_test.go
danbf: removed

internal/proxy/options.go
```diff
@@ -19,6 +19,7 @@ import (
 // Options are configuration options that can be set by Environment Variables
 // Port - int -  port to listen on for HTTP clients
 // ProviderURLString - the URL for the provider in this environment: "https://sso-auth.example.com"
+// ProxyProviderURLString - the internal URL for the provider in this environment: "https://sso-auth-int.example.com"
```
benjsto: If this field is truly intended as an internal URL, would it be clearer if the name reflected that? I struggle a little bit with the names `ProviderURLString` and `ProxyProviderURLString` being not meaningfully different as far as I can tell.

danbf: thought here is that the `sso_proxy` service is using this one, the other is for the `client`.  but then naming is not my strong suit

internal/proxy/providers/sso_test.go
```diff
@@ -271,6 +280,90 @@ func TestSSOProviderGetEmailAddress(t *testing.T) {
...
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			p := newSSOProvider()
+
+			var redeemServer *httptest.Server
+			var redeemServerInternal *httptest.Server
+			// set up redemption resource
+			if tc.RedeemResponseInternal != nil {
```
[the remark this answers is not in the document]
danbf: this is copied from the existing tests so i had not put much thought into it: https://github.com/buzzfeed/sso/blob/7898834043785d7b986fcbf22fb5a5271273b469/internal/proxy/providers/sso_test.go#L260-L266 

from what i can see it looks like the test cases are divided into two classes, those that don't produce a redeem response, and ones that should produce a valid redeem response which then gets checked for validity.  for this particular test, i'm testing to see that the code is hanlding the case where both the `RedeemURL` and `ProxyRedeemURL` both exist.  in that case the internal `ProxyRedeemURL` should be used for the redeem.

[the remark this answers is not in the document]
shrayolacrayon: Oh got it, I didn't see that we're still testing RedeemResponse above.

internal/proxy/options.go
```diff
@@ -215,10 +220,19 @@ func parseProviderInfo(o *Options) error {
 		return errors.New("provider-url must include scheme and host")
 	}
 
+	proxyProviderURL, err := url.Parse(o.ProxyProviderURLString)
+	if err != nil {
+		return err
+	}
+	if proxyProviderURL.Scheme == "" || proxyProviderURL.Host == "" {
+		return errors.New("proxyprovider-url must include scheme and host")
```
shrayolacrayon: nit: proxyprovider-url -> proxy provider url

danbf: fixed

Issue #90, opened 2018-10-15
Local make build failing due to file "statik/statik.go" already exists; use -f to overwrite


snebel29:
**The issue**
Running `make build` fails with Error 1

**To Reproduce**
1. Install sso from scratch following the instructions
```
$ go get github.com/buzzfeed/sso/cmd/...
$ cd $GOPATH/src/github.com/buzzfeed/sso
$ gpm install
$ make build
mkdir -p dist
go generate ./...
file "statik/statik.go" already exists; use -f to overwrite
internal/auth/static_files.go:38: running "/home/snebel/gopath/bin/statik": exit status 1
Makefile:8: recipe for target 'dist/sso-auth' failed
make: *** [dist/sso-auth] Error 1
```

**Expected behavior**
I would expect the make command to just build it without errors

**Desktop (please complete the following information):**
 - OS: 
```Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.5 LTS
Release:	16.04
Codename:	xenial
```
- Repository commit: `0519150fcf83dde5dcb077b4eff8689276b0150d`

**Additional context**
Adding `-f` flag to `internal/auth/static_files.go:38` seems to fix the issue
```
//go:generate $GOPATH/bin/statik -f -src=./static
```
```
$ make build
mkdir -p dist
go generate ./...
go build -o dist/sso-auth ./cmd/sso-auth
mkdir -p dist
go generate ./...
go build -o dist/sso-proxy ./cmd/sso-proxy
```
Although I don't have the full context on `statik` tool and how the CI is doing the building not to encounter any error, I think It'd be good to fix this to being able to do builds locally without errors.

Thanks

PR #91, opened 2018-10-15
Fix issue 90 Use -f on statik generate comment

snebel29:
Hi,
I think it's all detailed in this issue https://github.com/buzzfeed/sso/issues/90

I hope that helps


shrayolacrayon:
Thanks for opening @snebel29!

snebel29:
Thanks for reviewing @shrayolacrayon !

Issue #39, opened 2018-08-25
sso-auth: PROXY_ROOT_DOMAIN should be a required config value


mccutchen:
`PROXY_ROOT_DOMAIN` should be required configuration, because without it `sso-auth` will reject every redirect URL as invalid for reasons that are basically impossible to diagnose just from its error logs.

snebel29:
Since the PR was merged, I guess this is good to close

PR #92, opened 2018-10-15
Make PROXY_ROOT_DOMAIN required

snebel29:
## Problem
PROXY_ROOT_DOMAIN should be required as documented, as describe here https://github.com/buzzfeed/sso/issues/39

## Solution
Making that parameter required following the same patter as for others


shrayolacrayon:
Thank you for this contribution @snebel29!

PR #119, opened 2018-11-26
Godeps: remove unused dependencies

### LINKED ISSUE #10: remove use of go-options
**Is your feature request related to a problem? Please describe.**
We currently use go-options to allow us to either start sso-auth and sso-proxy via CLI or config file. Rather than having two different ways of starting these binaries, it seems like a good idea to stick to one way.

**Describe the solution you'd like**
* Run binaries using a config file.
   * Pro: More secure
    * Con: Will require more work.
* CLI Solution
   * Con: It is possible to leak sensitive secrets via CLI
   * Pro: Easier since we already use it.

mreiferson:
Fixes #10

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -1,8 +1,6 @@
 github.com/18F/hmacauth                  1.0.1
 gopkg.in/yaml.v2                         v2
 github.com/imdario/mergo                 v0.3.4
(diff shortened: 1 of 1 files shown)


mreiferson:
RFR @loganmeetsworld

PR #123, opened 2018-11-26
sso-proxy: internal host should apply to redeem, /refresh, /validate, /profile

benjsto:
## Problem

Dan's [PR](https://github.com/buzzfeed/sso/pull/88) added the ability to have an optional 'internal' host defined that `proxy` will use to connect to `auth`. However, that internal host was only applied on `/redeem` requests. Other cases where `proxy` makes a request to `auth`, like `/refresh`, `/validate`, or `/profile` should also use this internal host if it is provided.

## Solution

If the optional `PROVIDER_URL_INTERNAL` is set, use that host to connect to `auth` for /redeem, /refresh, /validate, and /profile.

### DIFF
--- a/internal/proxy/options.go
+++ b/internal/proxy/options.go
@@ -48,11 +48,11 @@ import (
 type Options struct {
 	Port int `envconfig:"PORT" default:"4180"`
 
-	ProviderURLString      string `envconfig:"PROVIDER_URL"`
-	ProxyProviderURLString string `envconfig:"PROXY_PROVIDER_URL"`
-	UpstreamConfigsFile    string `envconfig:"UPSTREAM_CONFIGS"`
-	Cluster                string `envconfig:"CLUSTER"`
-	Scheme                 string `envconfig:"SCHEME" default:"https"`
+	ProviderURLString         string `envconfig:"PROVIDER_URL"`
+	ProviderURLInternalString string `envconfig:"PROVIDER_URL_INTERNAL"`
+	UpstreamConfigsFile       string `envconfig:"UPSTREAM_CONFIGS"`
+	Cluster                   string `envconfig:"CLUSTER"`
+	Scheme                    string `envconfig:"SCHEME" default:"https"`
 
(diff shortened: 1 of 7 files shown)


internal/proxy/options.go
```diff
@@ -48,11 +48,11 @@ import (
 type Options struct {
 	Port int `envconfig:"PORT" default:"4180"`
 
-	ProviderURLString      string `envconfig:"PROVIDER_URL"`
-	ProxyProviderURLString string `envconfig:"PROXY_PROVIDER_URL"`
-	UpstreamConfigsFile    string `envconfig:"UPSTREAM_CONFIGS"`
-	Cluster                string `envconfig:"CLUSTER"`
-	Scheme                 string `envconfig:"SCHEME" default:"https"`
+	ProviderURLString         string `envconfig:"PROVIDER_URL"`
+	ProviderURLInternalString string `envconfig:"PROVIDER_URL_INTERNAL"`
```
danbf: I'd rather not make a breaking change here in the env-var name if possible even if we do it internally to make things clearer.  if we do change things, let's make sure we update the things like the quickstart:
https://github.com/buzzfeed/sso/blob/6db5495e8d4bdd50244e94cc2369517f37f74f38/quickstart/docker-compose.yml#L32

danbf:
@benjsto thanks much for finding these.  i like your solution in here neater then the one i pushed earlier.

internal/proxy/options.go
danbf: should we try to re-jigger the kubernetes quick start here https://github.com/buzzfeed/sso/blob/6db5495e8d4bdd50244e94cc2369517f37f74f38/quickstart/kubernetes/sso-proxy-deployment.yml#L26-L27 to enable the internal connection between sso-proxy->sso-auth ?

benjsto: yep, good idea - added.

benjsto:
Thanks @danbf for catching that the Host header on requests to `sso-auth` still need to use the _non-internal host, as that service specifically looks only for that Host header and will 404 without.

PR #128, opened 2018-11-29
sso-proxy: internal host should apply to redeem, /refresh, /validate, /profile

danbf:
local PR for: https://github.com/buzzfeed/sso/pull/123


benjsto:
Going to close this one since we've updated the original one with these changes.

PR #137, opened 2019-01-02
sso-proxy: use internal/pkg/sessions

### LINKED ISSUE #43: sso-proxy: use internal/pkg/sessions for session load and save logic

shrayolacrayon:
Changes all session store and csrf store logic to use internal/pkg/sessions. 

Fixes #43. 

RFR @buzzfeed/infra-security  @buzzfeed/sso-maintainers


internal/pkg/sessions/mock_store.go
```diff
@@ -41,10 +42,15 @@ func (ms *MockSessionStore) ClearSession(http.ResponseWriter, *http.Request) {
 
 // LoadSession returns the session and a error
 func (ms *MockSessionStore) LoadSession(*http.Request) (*SessionState, error) {
+	if ms.Session == nil {
+		return nil, http.ErrNoCookie
+	}
```
benjsto: pretty trivial but might be cleaner and more concise to to set ms.LoadError to http.ErrNoCookie here?

PR #138, opened 2019-01-04
Dev proxy

malizz:
## Problem
A production deployment does not allow failing verification, test permissions (that may not necessarily exist for them in prod).

## Solution

Simplifying development environment workflow by replacing the production proxy with a shim that reads its settings from a config file.


malizz:
Pull request opened by mistake.

PR #148, opened 2019-01-24
proxy: default allowed groups

jphines:
## Problem

We have the need to specify default allowed groups

## Solution

Allow us to specify default allowed groups via env config for every upstream.

## Notes
Needs some tests.


loganmeetsworld:
gonna need to add

```internal/proxy/options.go:118:16: unknown field 'DefaultGroups' in struct literal of type Options```

PR #149, opened 2019-01-24
auth: walk nested groups

jphines:
## Problem

We have nested google groups we need to unroll in-order to properly leverage them

## Solution

Recursively resolve nested groups up to 3 degrees.


internal/auth/providers/google_admin.go
```diff
@@ -52,7 +52,7 @@ func getAdminService(adminEmail string, credentialsReader io.Reader) *admin.Serv
 }
 
 // GetMembers returns the members of a google group
-func (gs *GoogleAdminService) GetMembers(groupName string) ([]string, error) {
+func (gs *GoogleAdminService) GetMembers(groupName string, tail int) ([]string, error) {
```
mccutchen: Super minor, but it's not obvious to me why this is called `tail`.  Seems like either `depth` (where we increment on recursion) or `maxDepth` (decrement) would be clearer.

internal/auth/providers/google_admin.go
```diff
@@ -109,6 +109,17 @@ func (gs *GoogleAdminService) GetMembers(groupName string) ([]string, error) {
 
 		for _, member := range r.Members {
 			members = append(members, member.Email)
```
shrayolacrayon: do we want to append the member.Email  if the member type is a group?

internal/auth/providers/google_admin.go
```diff
@@ -118,22 +137,19 @@ func (gs *GoogleAdminService) GetMembers(groupName string) ([]string, error) {
 	return members, nil
 }
 
-// GetGroups gets the groups that a user with a given email address belongs to.
-func (gs *GoogleAdminService) GetGroups(email string) ([]string, error) {
-	var groups []string
+// HasMember checks if the user has membership in the given groups
```
mccutchen: Nit: `HasMember` strikes me as an odd name for a function that returns a subset of the given groups that the given email is in.  Basically, I'd expect `HasMember` to return a bool.  What about something like `CheckMembership` instead?

(Also, it might be good to expand the docstring here to more accurately describe what this function is doing w/r/t returning a subset of the given groups.)

internal/auth/providers/google.go
```diff
@@ -338,9 +338,9 @@ func (p *GoogleProvider) ValidateGroupMembership(email string, allGroups []strin
 	groups := []string{}
 	var useGroupsResource bool
 
-	// if `allGroups` is empty use the groups resource
+	// if `allGroups` is empty, we return an empty list
 	if len(allGroups) == 0 {
-		return p.AdminService.GetGroups(email)
+		return []string{}, nil
```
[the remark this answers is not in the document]
jphines: We added this originally to ensure backward compatibility when the proxy didn't use to ask for explicit groups, but instead asked for membership of all groups -- will update the docstring.

internal/auth/providers/google_admin.go
jphines: I wanted to reflect the underlying API endpoint we're leveraging on the google side (which is a bool endpoint, but is nice because it does nested group resolution). 

I agree this seems to imply this would be a bool by name along -- not sure what would be better.

PR #154, opened 2019-02-07
Use Go 1.11 and go mod

sylr:
gpm is totally obsolete so let's move to Go 1.11 and use go mod which is future standard for managing Go dependencies.


sporkmonger:
Yeah, I've found `gpm` to be wildly frustrating to work with. I fully support this switch.

shrayolacrayon:
Thank you for this contribution @sylr! This is definitely an approach that we want to move towards, however one thing that we'd like to discuss separately is whether or not to check in the vendor'ed dependencies. For the time being to get this merged can you check out the dependencies in this PR?

sylr:
@shrayolacrayon I personally prefer to have the libs vendor'ed for two reason:

- First it's easier for people to build the project when they check it out the first time. They just have to `make build` and everything should for (minus the tools required like golint and statik). Also if someone mess up its repo (or even delete it) you can still compile.
- Second when you have libs pinned to a certain version you do not mess with your gopath which can have the same libs but with different versions for other projects.

jphines:
@sylr if you can rebase this branch -- we'd love to get it in.

sylr:
@jphines done.

jphines:
@sylr thanks for the contribution! 🥇

Issue #151, opened 2019-01-31
Long requests with longer timeout: SSO seems to close the connection


victornoel:
I have configured SSO in kubernetes in front of a service and behind nginx-ingress with proper timeout settings.
My service has requests that can be as long as 5 minutes.
The logs of sso-proxy says that the requests succeeded (ended in 200) but nginx-ingress says that the upstream (i.e., sso-proxy) prematurely closed the connection.

I don't believe it is a timeout that breaks the connection because sso-proxy seems to say that everything is ok and all the timeouts are configured for 10 minutes.

From the logs below, it looks like the problem is coming from SSO, but I'm not clear how to debug this further.

Logs from my service:
```
:03:07.669 [1c6c89ef6fb5b77d](0383ba8aa92c27d85abee89c0f59355c) INFO  c.b.a.s.Backend - Received: POST https://xxxxx.com/api/yyyyy
11:03:07.709 [1c6c89ef6fb5b77d](0383ba8aa92c27d85abee89c0f59355c) INFO  c.b.a.s.p.m.Stuff - Starting computing
11:07:30.982 [1c6c89ef6fb5b77d](0383ba8aa92c27d85abee89c0f59355c) INFO  c.b.a.s.p.m.Stuff - Finished computing
```

Logs from sso-proxy:
```
{"action":"proxy","http_status":200,"level":"info","msg":"","remote_address":"10.132.0.7","request_duration":236941.048871,"request_method":"POST","request_uri":"xxxxx.com/api/yyyyy","service":"sso-proxy","time":"2019-01-31 11:11:28.13111","user":"a@xxxxx.com","user_agent":"Mozilla/5.0 (X11; Linux x86_64; rv:65.0) Gecko/20100101 Firefox/65.0"}
```

Logs from nginx-ingress:
```
2019/01/31 11:03:07 [warn] 319#319: *3111 a client request body is buffered to a temporary file /tmp/client-body/0000000005, client: 10.4.1.1, server: xxxxx.com, request: "POST /api/yyyyy HTTP/2.0", host: "xxxxx.com", referrer: "https://xxxxx.com/page"
2019/01/31 11:07:31 [error] 319#319: *3111 upstream prematurely closed connection while reading response header from upstream, client: 10.4.1.1, server: xxxxx.com, request: "POST /api/yyyyy HTTP/2.0", upstream: "http://10.4.1.65:4180/api/yyyyy", host: "xxxxx.com", referrer: "https://xxxxx.com/page"
```

[...]

PR #163, opened 2019-03-01
proxy: set tcp write timeout to max timeout timeout value or default

jphines:
## Problem

In #151, a problem was explored where SSO Proxy writes out `200 OK` on long requests but nothing is written out to the client. We were able to trace this issue down to a default configuration we specify for the `http.Server` that would timeout tcp connections after the timeout window specified but before the upstream response was received.

However, because of a bug in the golang stdlib https://github.com/golang/go/issues/21389, when these writes occurred we were unable to easily diagnose the connection had closed and that the write was unsuccessful - no error is returned nor is there any useful log output. 

Unfortunately, there is no way to cleanly handle this error case to make debugging more useful without re-writing portions of the `http.Server`.

## Solution

For now, we can make this edge case a little less sharp by handling these timeouts more intuitively. We set the `WriteTimeout` for the `http.Server` to now be the max of the either the default value configured or the max upstream timeout config. 

## Notes

Many thanks to @victornoel for raising both the issue and the mitigation implemented here.

### DIFF
--- a/internal/proxy/options.go
+++ b/internal/proxy/options.go
@@ -32,7 +32,7 @@ import (
 // ClientSecret - The OAuth Client Secret
 // DefaultUpstreamTimeout - the default time period to wait for a response from an upstream
 // DefaultUpstreamTCPResetDeadline - the default time period to wait for a response from an upstream
-// TCPWriteTimeout - http server tcp write timeout
+// TCPWriteTimeout - http server tcp write timeout - set to: max(default value specified, max(upstream timeouts))
 // TCPReadTimeout - http server tcp read timeout
 // CookieName - name of the cookie
 // CookieSecret - the seed string for secure cookies (optionally base64 encoded)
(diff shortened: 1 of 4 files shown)


internal/proxy/options.go
```diff
@@ -200,6 +200,14 @@ func (o *Options) Validate() error {
 		}
 	}
 
+	if o.upstreamConfigs != nil {
+		for _, uc := range o.upstreamConfigs {
+			if uc.Timeout > o.TCPWriteTimeout {
+				o.TCPWriteTimeout = uc.Timeout
```
desimone: What if TCPWriteTimeout is set to 0 to support streaming?

jphines: I think users will have to implement this at their own peril for the moment. I think we'd have to implement an actually underlying idle tcp connection -- but might be a more involved change.

For the moment, this fix delivers tangible value to users.

PR #168, opened 2019-03-27
CI: remove requirement to pull in dependency source code changes 

Jusshersmith:
## Problem
All CI builds are currently failing due to updates to vendored dependencies being required. Specifically, we run git diff as part of our tests (https://github.com/buzzfeed/sso/blob/master/scripts/test) which fails after the CI process itself tries to update these dependencies. Due to the use of `go mod vendor` we currently need to pull in updated source code for these dependencies too.

## Solution
We don't want to have to update the source code for vendored dependencies on each build - instead this can be done intermittently after a certain time period or when we want to pull in updates from certain dependencies. This PR aims to take away that requirement.

Commit https://github.com/buzzfeed/sso/commit/0c2c289d7157ef88ccd5eab9780b5117e3e0dcb6 includes changes to `./scripts/test` and `Dockerfile` to prevent dependency source code being pulled down on each build, and to allow use of modules within the $GOPATH.

Commit https://github.com/buzzfeed/sso/commit/d1bfabeded3b0bd534120e6dad6e58102dba8adb includes changes to `go.mod` and `go.sum` from running `./scripts/test` and `make tools`. 

## Notes

One potential issue not addressed in this PR is that if any changes to `go.sum` and `go.mod` are caused by https://github.com/buzzfeed/sso/blob/master/Makefile#L18-L20, then the build will fail due to git diffs. Running `./scripts/test` does not make these updates, so currently `make tools` needs to be locally ran to pull in these changes.

Relevant PR: https://github.com/buzzfeed/sso/pull/166 (this was initially created to simply make the required dependency changes to unblock builds, however now should not be needed)


jphines:
This looks good to me.

scripts/test
```diff
@@ -11,18 +11,20 @@ if [ -n "$res" ]; then
   exit 1
 fi
 
-git diff --exit-code 
-if [ "$?" -gt "0" ]; then
-  exit 1
-fi
+# where necessary attempts to add/remove modules from go.mod and go.sum.
+echo "running go mod tidy ..."
+go mod tidy
```
jphines: Let's remove this blank line so it's clear the `git diff --exit-code` is running in response to the `go mod tidy` command.

PR #174, opened 2019-04-03
sso: okta provider MVP

Jusshersmith:
### Okta provider integration
This implements the MVP for adding Okta as an oauth provider for SSO.
Documentation on set up of Okta resources in a separate PR: https://github.com/buzzfeed/sso/pull/182

This PR includes:

- A new Okta provider (based off of the Google provider to keep them consistent where reasonable)
- A set of tests for the Okta provider. Again, initially based off of the Google provider tests however changed/added to where necessary.
- Changes to the options parsing layer to accommodate new settings required. (the only additional env var required that does not have a default is `okta_org_url` which is the URL of the org's okta domain (e.g. test.okta.com))
- Now that we have multiple possible providers, some changes to the `internal_util.go` and it's tests package to separate logic into respective provider packages.
- Minimum required changes to the google provider and google provider tests to work with the above changes to `internal_util.go`
- Minimum required changes to `provdider_default.go` to work with the changes to `internal_util.go`

One notable thing missing is any group caching functionality - this will also be shipped in a separate, soon-to-come PR.

### DIFF
--- a/internal/auth/options.go
+++ b/internal/auth/options.go
@@ -16,9 +16,10 @@ import (
 )
 
 // Options are config options that can be set by environment variables
-// RedirectURL 	string - the OAuth Redirect URL. ie: \"https://internalapp.yourcompany.com/oauth2/callback\
-// ClientID - 	string - the OAuth ClientID ie "123456.apps.googleusercontent.com"
+// RedirectURL - string - the OAuth Redirect URL. ie: \"https://internalapp.yourcompany.com/oauth2/callback\
+// ClientID - string - the OAuth ClientID ie "123456.apps.googleusercontent.com"
 // ClientSecret string - the OAuth Client Secret
+// OrgName - string - if using Okta as the provider, the Okta domain to use
 // ProxyClientID - string - the client id that matches the sso proxy client id
 // ProxyClientSecret - string - the client secret that matches the sso proxy client secret
 // Host - string - The host that is in the header that is required on incoming requests
@@ -42,8 +43,10 @@ import (
 // PassUserHeaders - bool (default true) - pass X-Forwarded-User and X-Forwarded-Email information to upstream
 // SetXAuthRequest - set X-Auth-Request-User and X-Auth-Request-Email response headers (useful in Nginx auth_request mode)
 // Provider - provider name
+// ProviderServerID - string - if using Okta as the provider, the authorisation server ID (defaults to 'default')
 // SignInURL - provider sign in endpoint
 // RedeemURL - provider token redemption endpoint
+// RevokeURL - provider revoke token endpoint
 // ProfileURL - provider profile access endpoint
 // ValidateURL - access token validation endpoint
 // Scope - Oauth scope specification
@@ -67,6 +70,8 @@ type Options struct {
 
 	GoogleAdminEmail         string `envconfig:"GOOGLE_ADMIN_EMAIL"`
 	GoogleServiceAccountJSON string `envconfig:"GOOGLE_SERVICE_ACCOUNT_JSON"`
+
+	OrgURL string `envconfig:"OKTA_ORG_URL"`
 
 	Footer string `envconfig:"FOOTER"`
 
@@ -93,9 +98,12 @@ type Options struct {
 	SetXAuthRequest    bool `envconfig:"SET_XAUTHREQUEST" default:"false"`
 
 	// These options allow for other providers besides Google, with potential overrides.
-	Provider       string `envconfig:"PROVIDER" default:"google"`
+	Provider         string `envconfig:"PROVIDER" default:"google"`
+	ProviderServerID string `envconfig:"PROVIDER_SERVER_ID" default:"default"`
+
 	SignInURL      string `envconfig:"SIGNIN_URL"`
 	RedeemURL      string `envconfig:"REDEEM_URL"`
+	RevokeURL      string `envconfig:"REVOKE_URL"`
 	ProfileURL     string `envconfig:"PROFILE_URL"`
 	ValidateURL    string `envconfig:"VALIDATE_URL"`
 	Scope          string `envconfig:"SCOPE"`
@@ -172,6 +180,13 @@ func (o *Options) Validate() error {
 		msgs = append(msgs, "missing setting: required-host-header")
 	}
 
+	if len(o.OrgURL) > 0 {
+		o.OrgURL = strings.Trim(o.OrgURL, `"`)
+	}
+	if len(o.ProviderServerID) > 0 {
+		o.ProviderServerID = strings.Trim(o.ProviderServerID, `"`)
+	}
+
 	o.redirectURL, msgs = parseURL(o.RedirectURL, "redirect", msgs)
 
 	msgs = validateEndpoints(o, msgs)
@@ -222,6 +237,7 @@ func (o *Options) Validate() error {
 func validateEndpoints(o *Options, msgs []string) []string {
 	_, msgs = parseURL(o.SignInURL, "signin", msgs)
 	_, msgs = parseURL(o.RedeemURL, "redeem", msgs)
+	_, msgs = parseURL(o.RevokeURL, "revoke", msgs)
 	_, msgs = parseURL(o.ProfileURL, "profile", msgs)
 	_, msgs = parseURL(o.ValidateURL, "validate", msgs)
 
@@ -246,13 +262,16 @@ func newProvider(o *Options) (providers.Provider, error) {
 	}
 
 	var err error
+
 	if p.SignInURL, err = url.Parse(o.SignInURL); err != nil {
 		return nil, err
 	}
 	if p.RedeemURL, err = url.Parse(o.RedeemURL); err != nil {
 		return nil, err
 	}
-	p.RevokeURL = &url.URL{}
+	if p.RevokeURL, err = url.Parse(o.RevokeURL); err != nil {
+		return nil, err
+	}
 	if p.ProfileURL, err = url.Parse(o.ProfileURL); err != nil {
 		return nil, err
 	}
@@ -277,6 +296,12 @@ func newProvider(o *Options) (providers.Provider, error) {
 		googleProvider.GroupsCache = cache
 		o.GroupsCacheStopFunc = cache.Stop
 		singleFlightProvider = providers.NewSingleFlightProvider(googleProvider)
+	case providers.OktaProviderName:
+		oktaProvider, err := providers.NewOktaProvider(p, o.OrgURL, o.ProviderServerID)
+		if err != nil {
+			return nil, err
+		}
+		singleFlightProvider = providers.NewSingleFlightProvider(oktaProvider)
 	default:
 		return nil, fmt.Errorf("unimplemented provider: %q", o.Provider)
 	}
--- a/internal/auth/providers/internal_util.go
+++ b/internal/auth/providers/internal_util.go
@@ -1,8 +1,6 @@
 package providers
 
 import (
-	"io/ioutil"
-	"net/http"
 	"net/url"
 
 	log "github.com/buzzfeed/sso/internal/pkg/logging"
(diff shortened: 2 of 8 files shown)


internal/auth/providers/internal_util.go
```diff
@@ -47,24 +48,45 @@ func stripParam(param, endpoint string) string {
...
-	if len(header) == 0 {
-		params := url.Values{"access_token": {accessToken}}
-		endpoint = endpoint + "?" + params.Encode()
-	}
 
-	req, err := http.NewRequest("GET", endpoint, nil)
-	if err != nil {
-		logger.Error(err, "token validation request failed")
-		return false
+	switch p.Data().ProviderName {
```
jphines: We should probably move this into a proper interface function.

```golang
func (p *Povider) ValidToken(accessToken) bool {
... provider specific logic here
}
```

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+		p.UserInfoURL = &url.URL{
+			Scheme: "https",
+			Host:   fmt.Sprintf("%s.oktapreview.com", strings.Trim(orgName, "\"")),
+			Path:   "/oauth2/default/v1/userinfo"}
+	}
+	// takes token as a URL query and returns active/inactive, and further info if active.
+	if p.ValidateURL.String() == "" {
+		p.ValidateURL = &url.URL{
+			Scheme: "https",
+			Host:   fmt.Sprintf("%s.oktapreview.com", strings.Trim(orgName, "\"")),
```
jphines: `oktapreview` we won't want to keep here

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+func NewOktaProvider(p *ProviderData, orgName string) (*OktaProvider, error) {
+	if orgName == "" {
+		return nil, errors.New("missing setting: okta-org-name")
+	}
+	p.ProviderName = "Okta"
+	// interact with resource owner and obtain an authorization grant
+	if p.SignInURL.String() == "" {
+		p.SignInURL = &url.URL{
+			Scheme: "https",
+			Host:   fmt.Sprintf("%s.oktapreview.com", strings.Trim(orgName, "\"")),
```
jphines: Should we leave these `oktapreview.com`'s in here?

internal/auth/options.go
```diff
@@ -93,10 +98,13 @@ type Options struct {
...
 
 	// These options allow for other providers besides Google, with potential overrides.
-	Provider       string `envconfig:"PROVIDER" default:"google"`
+	Provider string `envconfig:"PROVIDER" default:"google"`
+
 	SignInURL      string `envconfig:"SIGNIN_URL"`
 	RedeemURL      string `envconfig:"REDEEM_URL"`
+	RevokeURL      string `envconfig:"REVOKE_URL"`
 	ProfileURL     string `envconfig:"PROFILE_URL"`
+	UserInfoURL    string `envconfig:"USERINFO_URL"`
```
jphines: Is this different than `ProfileURL`? If so, what is the difference?

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+
+	userinfo, err := p.UserInfo(AccessToken)
+	if err != nil {
+		return "", err
+	}
+	if userinfo.EmailAddress == "" {
+		return "", errors.New("missing email")
+	}
+	if !userinfo.EmailVerified {
+		return "", fmt.Errorf("email %s not listed as verified", userinfo.EmailAddress)
```
jphines: Let's not use a templated error struct here, but just return a generic `email not verified` and potentially log the error?

I think this will be nicer on structure log systems.

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+	if len(allowedGroups) == 0 {
+		return []string{}, nil
+	}
+
+	userinfo, err := p.UserInfo(accessToken)
+	if err != nil {
+		return nil, err
+	}
+	if len(userinfo.Groups) == 0 {
+		return nil, fmt.Errorf("no group membership returned for %s", email)
```
jphines: ditto here re logging generic error message w/out email formatting

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+
+// UserInfo takes in an access token and sends a request to the /userinfo endpoint.
+// Conditionally returns nil response or a struct of specified claims.
+func (p *OktaProvider) UserInfo(AccessToken string) (*UserInfoResponse, error) {
+	// https://developer.okta.com/docs/api/resources/oidc/#userinfo
+	response := &UserInfoResponse{}
+	if AccessToken == "" {
+		return nil, ErrBadRequest
+	}
+	var bearer = "Bearer " + AccessToken
```
jphines: let's string format rather than string concat

```golang
bearer := fmt.Sprintf("Bearer %s", AccessToken)
```

I also typically prefer `:=` variable instantiation over `var x`.

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+
+// OktaProvider is an implementation of the Provider interface.
+type OktaProvider struct {
+	*ProviderData
+	StatsdClient *statsd.Client
+	cb           *circuit.Breaker
+	GroupsCache  groups.MemberSetCache
+}
+
+type UserInfoResponse struct {
```
jphines: Let's make this an inline anonymous struct, which appears to be stylistically more consistent with other parts of this package.

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+		if err != nil {
+			p.StatsdClient.Incr("provider.internal_error", tags, 1.0)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func jwtDecodeSegmentOkta(seg string) ([]byte, error) {
```
jphines: unused?

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+	"github.com/buzzfeed/sso/internal/pkg/sessions"
+	"github.com/datadog/datadog-go/statsd"
+)
+
+// OktaProvider is an implementation of the Provider interface.
+type OktaProvider struct {
+	*ProviderData
+	StatsdClient *statsd.Client
+	cb           *circuit.Breaker
+	GroupsCache  groups.MemberSetCache
```
jphines: Until we can properly add this, let's drop it.

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+			Host:   fmt.Sprintf("%s.oktapreview.com", strings.Trim(orgName, "\"")),
+			Path:   "/oauth2/default/v1/authorize"}
+		// to get a refresh token. see https://developer.okta.com/authentication-guide/tokens/refreshing-tokens/
+	}
+	// returns access tokens, ID tokens or refresh tokens
+	if p.RedeemURL.String() == "" {
+		p.RedeemURL = &url.URL{
+			Scheme: "https",
+			Host:   fmt.Sprintf("%s.oktapreview.com", strings.Trim(orgName, "\"")),
+			Path:   "/oauth2/default/v1/token"}
```
jphines: I wonder if we can also generic-ify these paths, they could vary based on the namely structure of the authentication server on the okta side, right? i.e. it could be a non-`default` authentication service?

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,416 @@
...
+		ExpiresIn   int64  `json:"expires_in"`
+	}
+
+	err = p.oktaRequest("POST", p.RedeemURL.String(), params, []string{"action", "refresh"}, nil, &response)
+	if err != nil {
+		return
+	}
+
+	token = response.AccessToken
+	expires = time.Duration(response.ExpiresIn) * time.Second
```
jphines: let's make sure this time math is correct

internal/auth/providers/provider_data.go
```diff
@@ -9,12 +9,14 @@ import (
 // necessary to implement the Provider interface.
 type ProviderData struct {
 	ProviderName       string
+	oktaDomain         string
```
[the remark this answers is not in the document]
Jusshersmith: Ah, good catch. I don't believe we need this anymore. It was included while adding a `BaseURL` which would act as the base to Okta provider URL's (`SignInURL`, RedeemURL` etc). That's been removed as well.

internal/auth/providers/okta.go
Jusshersmith: Yup. `default` may be replaced with the ID of the auth server. I'll modify to allow this!

internal/auth/options.go
Jusshersmith: Interesting! It hadn't really triggered that it was defined but not in use at all within the `auth` configs, only the `proxy` configs. I think yes, it makes sense for this to be used in place of `UserInfoURL`! I've removed.

internal/auth/providers/google.go
```diff
@@ -119,7 +119,18 @@ func (p *GoogleProvider) SetStatsdClient(statsdClient *statsd.Client) {
 
 // ValidateSessionState attempts to validate the session state's access token.
 func (p *GoogleProvider) ValidateSessionState(s *sessions.SessionState) bool {
-	return validateToken(p, s.AccessToken, nil)
+	if s.AccessToken == "" {
+		return false
+	}
+
+	params := url.Values{"access_token": {s.AccessToken}}
+	endpoint := p.ValidateURL.String() + "?" + params.Encode()
```
jphines: hmm this is interesting.

As written, this isn't particularly safe if the validate url already has query params added to it. Modifying ValidatURL inline is also not ideal since it's a shared object

I think what we want here is to copy the ValidateURL object and then mutate the query like so: 
```golang
endpoint := &net.URL{}
reflect.DeepCopy(p.ValidateURL, endpoint)
q := endpoint.Query()
q.Add("access_token", s.AccessToken)
endpoint.RawQuery = q.Encode()
```

internal/auth/providers/okta.go
Jusshersmith: Clarified that in this case `UserInfoResponse` is passed around and used in multiple functions, which is why it was created as a global struct like so. Possible the logic that uses it could be altered to better match other examples in the package but for now I think it's alright to leave it like so?

internal/auth/providers/google.go
jphines: we could simplify this a bit by doing a simple pointer shallow copy also fwiw
```golang
var endpoint url.Url
endpoint = *p.ValidateURL
q := endpoint.Query()
q.Add("access_token", s.AccessToken)
endpoint.RawQuery = q.Encode()
```

internal/auth/providers/okta_test.go
```diff
@@ -0,0 +1,430 @@
...
+		t.Run(expected.name, func(t *testing.T) {
+			p := newOktaProvider(expected.providerData, t)
+			if p == nil {
+				t.Errorf("Okta provider was nil")
+			}
+			if p.Data().ProviderName != "Okta" {
+				t.Errorf("expected provider name Okta, got %s", p.Data().ProviderName)
+			}
+			if p.Data().SignInURL.String() != expected.signInURL {
+				log.Printf("expected %s", expected.signInURL)
```
jphines: instead of using the stdlib logger, we should use the test logger.

See https://golang.org/pkg/testing/#T.Logf

This makes it easier to work with limiting our log output for tests to show results where it matters.

internal/auth/providers/okta_test.go
```diff
@@ -0,0 +1,430 @@
...
+				log.Printf("got %s", p.Data().SignInURL.String())
```
jphines: Another little trick I've learned along the way is to line up the print statements so they are easier to visually parse for inconsistencies. In addition, I'll use `%q` to print quoted strings which makes it easier to see what space issues

For example:
```golang
log.Printf("expected %s", expected.signInURL)
log.Printf("got %s", gotURL)
```
which prints:
```
expected https://some.tricky.url.to.visually.parse.com/
got https://some.tricky.url.to.visualy.parse.com/ 
```
vs:
```golang
log.Printf("want %q", expected.signInURL)
log.Printf(" got %q", gotURL)
```
which prints:
```
want 'https://some.tricky.url.to.visually.parse.com/'
 got 'https://some.tricky.url.to.visualy.parse.com/ '
```

internal/auth/providers/okta.go
```diff
@@ -0,0 +1,423 @@
...
+			return nil, errors.New("missing setting: provider_server_id")
+		}
+	}
+
+	p.ProviderName = "Okta"
+	// interact with resource owner and obtain an authorization grant
+	if p.SignInURL.String() == "" {
+		p.SignInURL = &url.URL{
+			Scheme: "https",
+			Host:   fmt.Sprintf("%s.okta.com", strings.Trim(orgName, "\"")),
```
jphines: Let's move this line to a higher scope rather than repeating on every url construction block, same with `providerServerID` and `Scheme`.

```golang
scheme := "https"
host := fmt.Sprintf("%s.okta.com", strings.Trim(orgName, "\"")),
providerServerID := strings.Trim(providerServerID, "\"")
. . .
if p.SignInURL.String() == "" {
    p.SignInURL = &url.URL{
        Scheme: scheme,
        Host:   host,
        Path:   fmt.Sprintf("/oauth2/%s/v1/authorize", providerServerID),
    }
}
```

Secondly, why do we need to trim `"` from all these strings? It's a little weird to be doing at the provider module level. We should be stripping any lingering `"` at the options parse layer.

jphines:
@Jusshersmith this is looking really good, just a few nits about testing and arg validation!

internal/auth/providers/okta_test.go
Jusshersmith: This is a great idea - incredibly useful (yet simple)! 💯

internal/auth/providers/okta_test.go
Jusshersmith: Good point! 👍

internal/auth/providers/google.go
Jusshersmith: This is a great catch - thank you for the detailed code snippets!

internal/auth/providers/okta.go
Jusshersmith: 👍 I also moved the string trimming to the options parse layer.

internal/auth/providers/okta.go
Jusshersmith: As far as I can tell, this looks correct! But let me know if there's anything in particular you're concerned about.

PR #180, opened 2019-04-17
providers: support different providers per upstream

jphines:
## Problem

We want to start some larger restructuring within sso to allow upstreams to specify identity providers. 

This has several use cases:
* Support a slow transition from one identity provider
* Support legacy or otherwise ❄️ applications with unique identity requirements
* Support various degrees of scope permissions with separate providers
* Support multiple logical organizations within a single sso deployment

To begin, we want to move the provider object to be within the hierarchy of the upstream router itself. 

## Solution
Support multiple identity providers.


internal/proxy/options.go
```diff
@@ -189,10 +194,14 @@ func (o *Options) Validate() error {
...
-			Timeout:       o.DefaultUpstreamTimeout,
-			ResetDeadline: o.DefaultUpstreamTCPResetDeadline,
+			AllowedGroups:     o.DefaultAllowedGroups,
+			Timeout:           o.DefaultUpstreamTimeout,
+			ResetDeadline:     o.DefaultUpstreamTCPResetDeadline,
+			PassAccessToken:   o.PassAccessToken,
+			SkipAuthPreflight: o.SkipAuthPreflight,
+			ProviderSlug:      o.DefaultProviderSlug,
 		}
+		fmt.Printf("default upstream config opts: %v\n\n", defaultUpstreamOptionsConfig)
```
jphines: DELETE

internal/proxy/oauthproxy.go
```diff
@@ -602,37 +599,53 @@ func (p *OAuthProxy) isXHR(req *http.Request) bool {
...
 func (p *OAuthProxy) SignOut(rw http.ResponseWriter, req *http.Request) {
+	tags := []string{"action:sign_out"}
+
 	p.sessionStore.ClearSession(rw, req)
 
+	route, ok := p.router(req)
+	if !ok {
+		// this shouldn't happen since we've already matched the host once on this request
+		tags = append(tags, "error:unknown_host")
+		p.StatsdClient.Incr("application_error", tags, 1.0)
```
Jusshersmith: I might not quite be thinking about the recorded metrics/stats correctly, but I see in the tags here we include the 'action' as well as the  'error' as opposed to in the `p.UnknownHost` function called below we only include the 'error'. However, should we only increment the `application_error` metric once per 'UnknownHost' so as to not mislead the metric data?

internal/proxy/options.go
```diff
@@ -80,11 +84,9 @@ type Options struct {
...
 	CookieHTTPOnly bool          `envconfig:"COOKIE_HTTP_ONLY"`
 
-	PassAccessToken bool `envconfig:"PASS_ACCESS_TOKEN" default:"false"`
-
-	// These options allow for other providers besides Google, with potential overrides.
-	Provider string `envconfig:"PROVIDER" default:"google"`
-	Scope    string `envconfig:"SCOPE"`
+	// TODO: What should we do with scope? Right now it doesn't seem to make much sense since SSO is fairly opiniated
+	// in what it needs and asks for from identity providers. Do we need to allow this to overridden?
+	Scope string `envconfig:"SCOPE"`
```
Jusshersmith: Hmm..I'm not sure if it's something that would be used in the short term...maybe at some point in the future if further restructuring is done, in which case it could always be re-added.

But right now, I don't think overriding the scope brings about any benefits (Overriding and _adding_ an additional scope **or** _reducing_ the scopes wouldn't necessarily change anything functionality wise, but could easily cause issues)?

I'm not sure if there's some edge case where it would be useful though 🤔

PR #181, opened 2019-04-23
Extend timeouts

sporkmonger:
We're using SSO proxy to sit in the middle of another auth component that blocks waiting for authentication to complete. We need a long timeout to allow the user to finish the authentication flow that SSO is mediating. The timeout needs to at least get us past p99 time-to-complete-authentication and 5 minutes seems to be approximately the sweet spot.

### DIFF
--- a/internal/proxy/oauthproxy.go
+++ b/internal/proxy/oauthproxy.go
@@ -185,8 +185,8 @@ func (t *upstreamTransport) getTransport() *http.Transport {
 		t.transport = &http.Transport{
 			Proxy: http.ProxyFromEnvironment,
 			DialContext: (&net.Dialer{
-				Timeout:   30 * time.Second,
-				KeepAlive: 30 * time.Second,
+				Timeout:   300 * time.Second,
+				KeepAlive: 300 * time.Second,
 				DualStack: true,
 			}).DialContext,
 			MaxIdleConns:          100,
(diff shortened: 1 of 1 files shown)


internal/proxy/oauthproxy.go
```diff
@@ -185,8 +185,8 @@ func (t *upstreamTransport) getTransport() *http.Transport {
 		t.transport = &http.Transport{
 			Proxy: http.ProxyFromEnvironment,
 			DialContext: (&net.Dialer{
-				Timeout:   30 * time.Second,
-				KeepAlive: 30 * time.Second,
+				Timeout:   300 * time.Second,
```
jphines: We should figure out a way to make this configurable. A default of 300 seconds (much like 30 seconds) is unlikely to be sufficient for everyone's needs.

internal/proxy/oauthproxy.go
```diff
@@ -216,6 +216,7 @@ func NewReverseProxy(to *url.URL, config *UpstreamConfig) *httputil.ReverseProxy
 			req.Host = to.Host
 		}
 	}
+	proxy.FlushInterval = 100 * time.Millisecond
```
jphines: I don't think we should supply this default since this is configurable per upstream.

sporkmonger: That can be dropped, yeah.

PR #183, opened 2019-04-25
Auth/multi providers

jphines:
## Problem

As a complement to https://github.com/buzzfeed/sso/pull/180, which introduces the ability for sso proxy to specify providers per upstream, this allows sso auth to front multiple providers by accepting `provider_slug` on query parameter either as a get parameter or a body parameter for post requests. 

## Notes

Despite having underlying support to supply multiple providers, we don't yet expose the ability to configure multiple providers. We will likely leverage https://github.com/spf13/viper to extend our ability to manage complex configuration.

### DIFF
--- a/cmd/sso-auth/main.go
+++ b/cmd/sso-auth/main.go
@@ -41,7 +41,13 @@ func main() {
 		return nil
 	}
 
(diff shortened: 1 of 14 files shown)


jphines:
Closing for #180

PR #184, opened 2019-04-30
providers: group caching mechanism for okta provider

Jusshersmith:
## Caching mechanism for groups returned by the Okta provider
- Includes a new module in the `groups` package: **internal/pkg/groups/localcache.go** for local caching of groups.
    - Tests included
- A new 'GroupCache' provider which wraps subsequent provider functions, including checks for cached entries using the above 'localcache' for the provider's `ValidateGroupMembership` function.
    - Tests included
- Adds a `statsdClient` to the provider interface (and adds/modifies the relevant provider function where necessary)
- New vendored depedency for https://godoc.org/golang.org/x/sync/syncmap

### DIFF
--- a/internal/auth/options.go
+++ b/internal/auth/options.go
@@ -38,6 +38,8 @@ import (
 // CookieHTTPOnly - bool - set httponly cookie flag
 // RequestTimeout - duration - overall request timeout
 // AuthCodeSecret - string - the seed string for secure auth codes (optionally base64 encoded)
+// GroupCacheProviderTTL - time.Duration - cache TTL for the group-cache provider used for on-demand group caching
+// GroupsCacheRefreshTTL - time.Duratoin - cache TTL for the groups fillcache mechanism used to preemptively fill group caches
 // PassHostHeader - bool - pass the request Host Header to upstream (default true)
 // SkipProviderButton - bool - if true, will skip sign-in-page to directly reach the next step: oauth/start
 // PassUserHeaders - bool (default true) - pass X-Forwarded-User and X-Forwarded-Email information to upstream
@@ -89,6 +91,7 @@ type Options struct {
 
 	AuthCodeSecret string `envconfig:"AUTH_CODE_SECRET"`
 
+	GroupCacheProviderTTL time.Duration `envconfig:"GROUP_CACHE_PROVIDER_TTL" default:"10m"`
 	GroupsCacheRefreshTTL time.Duration `envconfig:"GROUPS_CACHE_REFRESH_TTL" default:"10m"`
 	SessionLifetimeTTL    time.Duration `envconfig:"SESSION_LIFETIME_TTL" default:"720h"`
 
@@ -301,7 +304,10 @@ func newProvider(o *Options) (providers.Provider, error) {
 		if err != nil {
 			return nil, err
 		}
-		singleFlightProvider = providers.NewSingleFlightProvider(oktaProvider)
+		tags := []string{"provider:okta"}
+
+		groupsCache := providers.NewGroupCache(oktaProvider, o.GroupCacheProviderTTL, oktaProvider.StatsdClient, tags)
+		singleFlightProvider = providers.NewSingleFlightProvider(groupsCache)
 	default:
 		return nil, fmt.Errorf("unimplemented provider: %q", o.Provider)
 	}
@@ -334,14 +340,7 @@ func AssignStatsdClient(opts *Options) func(*Authenticator) error {
 			"statsd client is running")
 
 		proxy.StatsdClient = StatsdClient
-		switch v := proxy.provider.(type) {
-		case *providers.GoogleProvider:
-			v.SetStatsdClient(StatsdClient)
-		case *providers.SingleFlightProvider:
-			v.AssignStatsdClient(StatsdClient)
-		default:
-			logger.Info("provider does not have statsd client")
-		}
+		proxy.provider.SetStatsdClient(StatsdClient)
 		return nil
 	}
 }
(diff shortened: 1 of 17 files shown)


internal/auth/options.go
```diff
@@ -337,6 +339,8 @@ func AssignStatsdClient(opts *Options) func(*Authenticator) error {
 		switch v := proxy.provider.(type) {
 		case *providers.GoogleProvider:
 			v.SetStatsdClient(StatsdClient)
```
jphines: We should probably just add this to the interface at this point so we don't have to do the type switching here.

internal/auth/providers/singleflight_middleware.go
```diff
@@ -124,7 +126,14 @@ func (p *SingleFlightProvider) ValidateGroupMembership(email string, allowedGrou
 	sort.Strings(allowedGroups)
 	response, err := p.do("ValidateGroupMembership", fmt.Sprintf("%s:%s", email, strings.Join(allowedGroups, ",")),
 		func() (interface{}, error) {
-			return p.provider.ValidateGroupMembership(email, allowedGroups, accessToken)
+			switch v := p.provider.(type) {
```
jphines: hmm this is really not what I anticipated us doing. the `GroupsCache` should be fulfilling the entire interface, instead of just the one method.

internal/auth/providers/groups_cache.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/datadog/datadog-go/statsd"
+)
+
+type Cache interface {
+	Get(keys string)
+	Set(entries []string)
+	Purge(keys []string)
+}
+
+type GroupCache struct {
```
jphines: This struct isn't fulfilling the interface, which I think we talked about this doing. 

Let's add a compile time check

```golang
// compile time interface check
var _ Provider =&GroupCache{}
```

And we can then fulfill the interface in one of two ways:

### 1. Embed the provider.
We can embed the provider in the struct by doing the following:

```golang
type GroupCache struct {
    metrics *statsd.Client
    cache *groups.LocalCache
    
    Provider
}
```

### 2. Stub out the rest of the interface
We supply stub methods for each of the interface methods which simply just call the underlying provider
```golang
func (p* GroupCache) OtherInterfaceFuncs( . . .{
    return p.provider.OtherInterFaceFuncs(...
}
```

Generally, 2 is preferred as it is more explicit and highlights the complexities of changing the interface if necessary. 1 is a convenient and can be useful if the changes you are making are minimal.

internal/auth/providers/groups_cache.go
```diff
@@ -0,0 +1,59 @@
...
+	return &GroupCache{
+		metrics:  statsdClient,
+		provider: provider,
+		cache:    groups.NewLocalCache(ttl, statsdClient, tags),
+	}
+}
+
+// ValidateGroupMembership wraps the provider's ValidateGroupMembership around calls to check local cache for group membership information.
+func (p *GroupCache) ValidateGroupMembership(email string, allowedGroups []string, accessToken string) ([]string, error) {
+	groupMembership, err := p.cache.Get(email)
```
jphines: We also don't check this err!

internal/auth/providers/groups_cache.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// If the user's group membership is not in cache, or the passed list of 'AllowedGroups'
+	// differs from the cached entry, call and return the groups from p.Provider.ValidateGroupMembership.
+	validGroups, err := p.provider.ValidateGroupMembership(email, allowedGroups, accessToken)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create and cache an Entry and return the valid groups.
+	entry := groups.Entry{
```
jphines: Do we have a TTL for this entry?

internal/auth/providers/groups_cache.go
```diff
@@ -0,0 +1,59 @@
...
+		metrics:  statsdClient,
+		provider: provider,
+		cache:    groups.NewLocalCache(ttl, statsdClient, tags),
+	}
+}
+
+// ValidateGroupMembership wraps the provider's ValidateGroupMembership around calls to check local cache for group membership information.
+func (p *GroupCache) ValidateGroupMembership(email string, allowedGroups []string, accessToken string) ([]string, error) {
+	groupMembership, err := p.cache.Get(email)
+	if reflect.DeepEqual(groupMembership.UserGroupData.AllowedGroups, allowedGroups) && len(groupMembership.UserGroupData.MatchedGroups) > 0 {
```
jphines: This is ALOT of logic on this one line, can we split this to multiplel lines? The `reflect.DeepEqual` will only return true if each entry in the structs are equal so we will want to sort the output before we compare them. We can store the output sorted in the cache and just sort the passed groups or sort both on demand.

internal/auth/providers/groups_cache.go
Jusshersmith: It's currently passed in when creating the local cache, so it's the same TTL for each key. 
```
return &GroupCache{
		metrics:  statsdClient,
		provider: provider,
		cache:    groups.NewLocalCache(ttl, statsdClient, tags),
	}
}
```
Would you suggest setting the TTL with each 'entry', so it can differ with each? (key-by-key basis, rather than whole cache-level?)

internal/auth/providers/groups_cache.go
Jusshersmith: Ah, yes! Sorting the groups is something I had intended to add when looking back through this but clearly neglected to do so...so thank you for pointing this out!

internal/auth/providers/singleflight_middleware.go
```diff
@@ -20,6 +20,7 @@ var (
 )
 
 // ErrUnexpectedReturnType is an error for an unexpected return type
+// ErrUnknownIdentityProvider is for when an unknown Identity Provider is given.
```
jphines: I think we can also remove this err

internal/auth/providers/groups_cache.go
Jusshersmith: Thank you for the explanations + examples!

PR #186, opened 2019-05-06
docker: don't run as root

cameronattard:
## Problem

It is best practice to run containers as a non-root user. Currently, the `sso` container runs as root.

## Solution

Create a non-root user, `sso`, in the image and configure the container to run as this user using the `USER` directive.


mreiferson:
Hi @cameronattard, thanks for this!

FYI I pushed up a whitespace at EOF fix, waiting for the build to test locally.

mreiferson:
(validated this with quickstart)

PR #187, opened 2019-05-06
sso_proxy: refactor hostmux/reverseproxy/tests

jphines:
## Problem

SSO up to this point has been structured to share a single provider and oauthproxy instance. This OAuthProxy object did the complex routing, housing various routing types, differentiated different upstreams for authorization, and managed the complexity of switching between various context states.

This, however, adds undue complexity and makes it harder to add more distinct features.

## Solution

This is a refactor of the `OAuthProxy` code base which emphasizes:

* Adds smaller, a targeted package for routing, `hostmust`
* Now implements a 1:1:1:1 relationship to `OAuthProxy`, `UpstreamConfig`, `Provider`, and `ReverseProxy` as opposed to `1:N:1:N` previously.
* Cleans up reverse proxy logic and splits it out into it's own testable component
* Cleans up test overhead complexity by simplifying `OAuthProxy` startup and adding new func option helpers.
* Adds new `SSOProxy` parent object to house routing, `/ping` and instantiation of many `OAuthProxy` objects

PR #188, opened 2019-05-07
sso-proxy: refactor reverse proxy into separate file w/ tests

jphines:
## Problem

Refactor reverse proxy and upstream into separate file w/ tests. Cleans up some of the reverse proxy logic.

This should make it easier to test reverse proxy specific features/changes in the future and help isolate the reverse proxy from the authentication/authorization portions of sso.

A split out from https://github.com/buzzfeed/sso/pull/187


internal/proxy/reverse_proxy.go
```diff
@@ -0,0 +1,212 @@
...
+			TLSHandshakeTimeout:   10 * time.Second,
+			TLSClientConfig:       &tls.Config{InsecureSkipVerify: t.insecureSkipVerify},
+			ExpectContinueTimeout: 1 * time.Second,
+		}
+	}
+
+	return t.transport
+}
+
+func NewUpstreamReverseProxy(config *UpstreamConfig, signer *RequestSigner) (http.Handler, error) {
```
shrayolacrayon: are we linting in our CI?

internal/proxy/reverse_proxy.go
```diff
@@ -0,0 +1,212 @@
...
+		Transport: &upstreamTransport{
+			resetDeadline:      config.ResetDeadline,
+			insecureSkipVerify: config.TLSSkipVerify,
+		},
+		ModifyResponse: func(resp *http.Response) error {
+			// DRAGONS: This helps implement special behavior regarding security headers
+			// We do not allow upstreams to override security headers and this is logic
+			// is required to delete the headers from the response before we add new headers in
+			// the response writer for all requests. If we add them here, they will only be added
+			// to authenicated requests.
```
shrayolacrayon: authenticated*

internal/proxy/reverse_proxy.go
jphines: Yes, https://github.com/buzzfeed/sso/blob/master/scripts/test#L28

Is there something here that would break linting?

internal/pkg/logging/logging.go
```diff
@@ -36,6 +40,43 @@ func (l *LogEntry) Fields() map[string]interface{} {
 	return l.logger.Data
 }
 
+// Convience Wrappers around logging actions
+
+// Info wraps the logrus Info function
+func Info(args ...interface{}) {
+	defaultLogger.logger.Info(args...)
+}
+
```
Jusshersmith: I can see the difference between this set of logging wrappers and the below ones, but I'm a little bit unclear around the motivation here - do we need both sets of wrappers, or can they be merged together (with either solution around instantiating a new logger being the predominant one)?

internal/proxy/middleware.go
```diff
@@ -19,22 +19,21 @@ var securityHeaders = map[string]string{
...
+func setHeaders(h http.Handler, headers map[string]string) http.Handler {
 	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		for key, val := range securityHeaders {
+		for key, val := range headers {
 			rw.Header().Set(key, val)
 		}
 		h.ServeHTTP(rw, req)
 	})
+
+}
```
jphines: Needs newline

internal/proxy/reverse_proxy.go
```diff
@@ -0,0 +1,212 @@
...
+			for key := range securityHeaders {
+				resp.Header.Del(key)
+			}
+			return nil
+		},
+	}
+
+	if config.FlushInterval != 0 {
+		reverseProxy.FlushInterval = config.FlushInterval
```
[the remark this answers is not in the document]
jphines: We don't `0` specifies no periodic flushing according to https://golang.org/pkg/net/http/httputil/ which is fine default behavior imho

PR #195, opened 2019-05-10
sso_auth: add identity provider slug to url path and mux

jphines:
## Problem

As we move towards supporting multiple providers within `sso-auth`, we need to provide unique callback urls for each identity provider.

## Solution

We add more provider slug prefixes to url paths and supply a default provider to support unique callback urls as well as preserve backward's compatibility

## Notes
* Creates a new `AuthenticatorMux` struct to house the mux mechanism for multiple `Authenticators`
* Move initialization from `main.go` to `NewAuthenticatorMux`
* Make `newStatsdClient` public to call from initialization from `main.go`
* Add new `Slug` struct field to `ProviderData` to differentiate between providers.

### DIFF
--- a/internal/auth/options.go
+++ b/internal/auth/options.go
@@ -7,13 +7,15 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"reflect"
 	"strings"
 	"time"
 
 	"github.com/buzzfeed/sso/internal/auth/providers"
 	"github.com/buzzfeed/sso/internal/pkg/groups"
-	log "github.com/buzzfeed/sso/internal/pkg/logging"
+
+	"github.com/datadog/datadog-go/statsd"
 	"github.com/spf13/viper"
 )
 
@@ -38,6 +40,7 @@ import (
 // CookieRefresh - duration - refresh the cookie after this duration default 0
 // CookieSecure - bool - set secure (HTTPS) cookie flag
 // CookieHTTPOnly - bool - set httponly cookie flag
+// DefaultProvider - string - specify the default provider
 // RequestTimeout - duration - overall request timeout
 // AuthCodeSecret - string - the seed string for secure auth codes (optionally base64 encoded)
 // GroupCacheProviderTTL - time.Duration - cache TTL for the group-cache provider used for on-demand group caching
@@ -104,7 +107,10 @@ type Options struct {
 
 	// These options allow for other providers besides Google, with potential overrides.
 	Provider         string `mapstructure:"provider"`
+	ProviderSlug     string `mapstructure:"provider_slug"`
 	ProviderServerID string `mapstructure:"provider_server_id"`
+
+	DefaultProvider string `mapstructure:"default_provider"`
 
 	SignInURL      string `mapstructure:"signin_url"`
 	RedeemURL      string `mapstructure:"redeem_url"`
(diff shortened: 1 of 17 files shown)


internal/auth/options.go
```diff
@@ -326,13 +330,22 @@ func AssignProvider(opts *Options) func(*Authenticator) error {
...
+}
+
 // AssignStatsdClient is function that takes in an Options struct and assigns a statsd client
 // to the proxy and provider.
 func AssignStatsdClient(opts *Options) func(*Authenticator) error {
 	return func(proxy *Authenticator) error {
 		logger := log.NewLogEntry()
 
-		StatsdClient, err := newStatsdClient(opts.StatsdHost, opts.StatsdPort)
+		StatsdClient, err := NewStatsdClient(opts.StatsdHost, opts.StatsdPort)
```
jphines: Instead of creating a new client, can we pass in the old one?

Create a new func:
```
SetStatsdClient(statsdClient *statsd.Client) func ....
```

and do this logic there?

Jusshersmith: I like this idea, though I also like the pattern of just passing in a relevant set of options (Would you pass the previously created statsdclient through the `NewAuthenticatorMux` function for example?) and having it create/assign the necessary components (e.g. the statsdclient). 

I think however creating and storing the statsdclient in the opts struct might be a strange direction. I don't feel particularly strongly regardless - just some general thoughts.

internal/auth/options.go
```diff
@@ -102,8 +102,11 @@ type Options struct {
 
 	// These options allow for other providers besides Google, with potential overrides.
 	Provider         string `envconfig:"PROVIDER" default:"google"`
+	ProviderSlug     string `envconfig:"PROVIDER_SLUG" default:"google"`
```
Jusshersmith: Are there any foreseeable situations where the provider slug would be different from the provider, or is this mostly a logical/organisational change?

internal/auth/options.go
```diff
@@ -326,13 +330,22 @@ func AssignProvider(opts *Options) func(*Authenticator) error {
 	}
 }
 
+// SetProvider is a function that takes an Options struct and assigns the
+// appropriate provider to the proxy.
+func SetProvider(provider providers.Provider) func(*Authenticator) error {
```
Jusshersmith: I think the comment needs to be slightly altered with it's inputs

internal/auth/options.go
```diff
@@ -388,19 +393,14 @@ func SetProvider(provider providers.Provider) func(*Authenticator) error {
 
 // AssignStatsdClient is function that takes in an Options struct and assigns a statsd client
 // to the proxy and provider.
-func AssignStatsdClient(opts *Options) func(*Authenticator) error {
+func AssignStatsdClient(statsdClient *statsd.Client) func(*Authenticator) error {
 	return func(proxy *Authenticator) error {
```
shrayolacrayon: unrelated nit: can we change this variable name to be authenticator rather than proxy?

internal/auth/options.go
jphines: One could think of a future where we would want to run multiple google identity providers, maybe one for division A and another for division B. In such a world, we would need a logical identifier to separate the two. Thus, we can leverage `ProviderSlug` which is a pure client-side string used to uniquely identify a specific google identity provider.

internal/auth/options.go
jphines: Done.

internal/auth/options.go
jphines: Done

PR #196, opened 2019-05-13
sso: introduce Viper for sso config management

Jusshersmith:
## Problem
Although config management within SSO is fairly straightforward at the moment, it is also unorganised and doesn't leave much room for extension. Particularly in the sense of multi-provider workflows or for storing the config elsewhere (e.g. remote config). Fulfilling these workflows right now config wise is tricky.

## Solution

Implement a new set of methods to manage config variables using a new package dependency such as [Viper](https://github.com/spf13/viper).
This PR currently introduces back-end changes that allow us to achieve this goal, while retaining functionality and flow (the same things are passed in, and the same things are returned). 

Near-future PR's will include further changes to mould the usage of Viper to better fit our needs.


internal/auth/options.go
```diff
@@ -129,20 +131,82 @@ type SignatureData struct {
...
+
+	err := viper.Unmarshal(&opts)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decode env vars into options struct")
+	}
+	return &opts, nil
+}
+
+// bindAllOptVars takes in a struct with tags and uses the tag values to bind env vars
+func bindAllOptVars(t reflect.Type, tag string) error {
```
Jusshersmith: Not necessarily required with name-spacing (setting prefix will allow a similar result)

internal/auth/authenticator_test.go
```diff
@@ -71,7 +71,7 @@ var testEncodedCookieSecret = "x7xzsM1Ky4vGQPwqy6uTztfr3jtm/pIdRbJXgE0q8kU="
 var testAuthCodeSecret = "qICChm3wdjbjcWymm7PefwtPP6/PZv+udkFEubTeE38="
 
 func testOpts(proxyClientID, proxyClientSecret string) *Options {
-	opts := NewOptions()
+	opts, _ := NewOptions()
```
jphines: My preference would be to check the err here if possible. In the case of these tests, let's pass a `t *testing.T` and fatal if these options aren't instantiated correctly.

internal/auth/options.go
```diff
@@ -129,20 +131,82 @@ type SignatureData struct {
...
+		err := viper.BindEnv(tagValue)
+		if err != nil {
+			return fmt.Errorf("Unable to bind env var: %q", tagValue)
+		}
+	}
+	return nil
+}
+
+// setDefaults sets config defaults for the default viper instance
+func setDefaults() {
```
jphines: my gut is to say we should take a `v *viper.Viper` here and use that to set defaults

internal/auth/options.go
```diff
@@ -129,20 +131,82 @@ type SignatureData struct {
...
-		CookieExpire:    time.Duration(168) * time.Hour,
-		CookieRefresh:   time.Duration(0),
-		SetXAuthRequest: false,
-		PassUserHeaders: true,
-		PassHostHeader:  true,
-		ApprovalPrompt:  "force",
-		RequestLogging:  true,
+// NewOptions returns new options with the below overrides
+func NewOptions() (*Options, error) {
+	overrides := map[string]interface{}{
```
jphines: It seems strange to me that we need both these as well as the defaults. Is there any way we could consolidate the two?

internal/auth/options.go
```diff
@@ -57,65 +59,65 @@ import (
 // StatsdPort - port where statsd client listens
 // StatsdHost - host where statsd client listens
 type Options struct {
-	RedirectURL       string `envconfig:"REDIRECT_URL" `
-	ClientID          string `envconfig:"CLIENT_ID"`
-	ClientSecret      string `envconfig:"CLIENT_SECRET"`
-	ProxyClientID     string `envconfig:"PROXY_CLIENT_ID"`
-	ProxyClientSecret string `envconfig:"PROXY_CLIENT_SECRET"`
+	RedirectURL       string `mapstructure:"redirect_url" `
```
jphines: Do we need these `mapstructure` to support the `_` from the env vars?

Jusshersmith: As I understand it, internally viper stores the env vars in a map and then uses https://github.com/mitchellh/mapstructure to decode into the struct, so the tags here need to be `mapstructure`, instead of `envconfig`

internal/auth/options.go
Jusshersmith: We don't need both of these - this was just in order to mimc the previous behaviour of having default settings, but also hard overrides. (unless I misunderstood that behaviour)

Something I wanted to bring up to question anyway as I wasn't completely sure of the motivation behind it. Very happy to consolidate if there isn't any need.

internal/auth/options.go
Jusshersmith: 👍 Creating a new viper instance in `NewOptions` and passing it around to relevant functions.

internal/auth/options.go
jphines: If it's easy, let's try to consolidate.

internal/auth/options.go
Jusshersmith: This will be addressed, if required, in a future PR. For now - it makes sense with the included changes.

Issue #197, opened 2019-05-16
Configure Proxy Cookie Name from Env Var


alexbakerdev:
**Is your feature request related to a problem? Please describe.**
Generally this is just making it possible to set the proxy CookieName via the environment.

One scenario where its useful is having multiple auth environments for a single cookie domain (like staging and production subdomains). E.g.

| Cookie Name | Domain |
|---|---|
| _sso_proxy_staging|  `staging.example.com` |
| _sso_proxy |  `example.com` |

Now the staging proxy can read a valid token from the cookies of the request 

**Describe the solution you'd like**
Make the cookie name configurable via an environment variable.

**Describe alternatives you've considered**
N/A

jphines:
Closed by #198

PR #198, opened 2019-05-16
Configure Proxy CookieName via Env Var

alexbakerdev:
## Problem

See #197 

## Solution

Add a COOKIE_NAME env var to the option struct

## Notes

It looks like there are already some tests around custom cookie name values, and I couldn't find any tests based around environment variables I couldn't find any, but if there are some tests needed do let me know!

### DIFF
--- a/internal/proxy/options.go
+++ b/internal/proxy/options.go
@@ -73,7 +73,7 @@ type Options struct {
 	TCPWriteTimeout time.Duration `envconfig:"TCP_WRITE_TIMEOUT" default:"30s"`
 	TCPReadTimeout  time.Duration `envconfig:"TCP_READ_TIMEOUT" default:"30s"`
 
-	CookieName     string
+	CookieName     string        `envconfig:"COOKIE_NAME" default:"_sso_proxy"`
 	CookieSecret   string        `envconfig:"COOKIE_SECRET"`
 	CookieDomain   string        `envconfig:"COOKIE_DOMAIN"`
 	CookieExpire   time.Duration `envconfig:"COOKIE_EXPIRE" default:"168h"`


alexbakerdev:
Just saw #196 which seems to make this invalid, will leave it open in case that PR won't be merged for some time.

jphines:
@alexbakerdev were you able to verify this functionality in your application? 

Per your example, 

| Cookie Name | Domain |
|---|---|
| _sso_proxy_staging|  `staging.example.com` |
| _sso_proxy |  `example.com` |


SSO sets cookies per domain (no wildcards) so I would expect the applications on each of these domains to not sent the `_sso_proxy` cookie by the browser. This existing application behavior moderately negates the usefulness of this change as described in https://github.com/buzzfeed/sso/pull/197

alexbakerdev:
Yup its working for our company :D

alexbakerdev:
Ah yup, as you say by default the cookie is set per domain however we had an interesting use case where we had multiple backend apis that were exposed on different subdomains.
I.e. `frontend.domain.com`, `api.domain.com`

Setting the cookie per domain in this case causes issues; since the proxy needs to initiate the authorisation redirect flow for each domain, making a CORS request to `api.domain.com` (when you don't already have a valid cookie) the proxy will redirect to the auth server which will fail, as it doesn't support cors. 

We solved this by setting the cookie domain to `domain.com` so the same cookie is used for each of its subdomains, but this caused issues because of the nested nature of the staging sub domain.

I get its a super weird use case, but I feel like there could be other use cases where environments may share a domain for whatever reason, and a similar problem will arise.

jphines:
@alexbakerdev Thanks for the write-up! This makes more sense in given the context!

PR #199, opened 2019-05-22
sso: enable golangci-lint

jphines:
## Problem

We have a limited number of linters and other static code analyzers. We should add more to improve our codebase and make review better and easier.

## Solution

`golangci-lint` https://github.com/golangci/golangci-lint is the newest replacement for `gometalinter` which has been the staple multi-lint tool. This PR adds support for `golangci-lint` as well as fixes the codebase so it runs cleanly.

## Notes
Default linters include:
* deadcode
* errcheck
* gosimple
* govet
* ineffassign
* staticcheck
* structcheck
* typecheck
* unused:
* varcheck

We also enable additional linters:
* gofmt
* golint
* gocritic
* gosec
* misspell

We can add other linters in the future but I'd like to focus on getting the defaults

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,10 +15,6 @@ jobs:
           name: get tools
           command: make tools
       - run:
-          name: copy source
-          command: |
-            mkdir bin
-      - run:
           name: run lint and tests for both services
           command: |
             scripts/test
--- a/scripts/test
+++ b/scripts/test
@@ -3,13 +3,8 @@ set -e
 
 cd $(dirname "$BASH_SOURCE[0]")/..
 
-echo "running go fmt ..."
-res=$(go fmt ./...)
-if [ -n "$res" ]; then
-  echo "$res"
-  echo "gofmt failed..."
-  exit 1
-fi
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -14,10 +14,6 @@ jobs:
       - run:
           name: get tools
           command: make tools
-      - run:
-          name: copy source
-          command: |
-            mkdir bin
```
jphines: This doesn't make much sense in context (it doesn't copy source!)

scripts/test
```diff
@@ -3,13 +3,8 @@ set -e
...
 
-echo "running go fmt ..."
-res=$(go fmt ./...)
-if [ -n "$res" ]; then
-  echo "$res"
-  echo "gofmt failed..."
-  exit 1
-fi
+echo "running golangci-lint ..."
+golangci-lint run -E gofmt -E golint -E gocritic -E gosec -E misspell
```
jphines: This should move into a config file so running `golangci-lint` locally is consistent with what is run in CI

PR #200, opened 2019-05-23
sso_proxy: don't set timeout when flush interval is set

Jusshersmith:
## Problem

`http.TimeoutHandler` doesn't support request flushing, resulting in any flush interval set being ignored if a timeout is also set (we set a default timeout, but not flush interval, so right now this problem always occurs if a flush interval is set).

https://golang.org/src/net/http/server.go?s=98830:98898#L3129

## Solution

If a flush interval is set in the upstream config, return the handler without setting a timeout.

### DIFF
--- a/internal/proxy/reverse_proxy_test.go
+++ b/internal/proxy/reverse_proxy_test.go
@@ -159,28 +159,28 @@ func TestNewRewriteReverseProxy(t *testing.T) {
 		expectedStatus int
 	}{
 		{
-			name:           "tls true skip verify false preserve host false",
+			name:           "tls=true, skipVerify=false, preserveHost=false",
 			useTLS:         true,
 			skipVerify:     false,
 			preserveHost:   false,
 			expectedStatus: 502,
 		},
 		{
-			name:           "tls true skip verify true preserve host false",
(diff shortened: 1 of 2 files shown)


internal/proxy/reverse_proxy_test.go
```diff
@@ -484,6 +484,16 @@ func TestTimeoutHandler(t *testing.T) {
 			ExpectedStatusCode: 200,
 			ExpectedBody:       "OK",
 		},
+		{
+			name: "does not time out as flush interval is set",
+			config: &UpstreamConfig{
+				Timeout:       time.Duration(100) * time.Millisecond,
```
jphines: Taking a deep look at this PR there are some issues with these tests:

1. We want to set a duration below `50` since that's how long we need to wait for the server below to throw an error. Any higher and we wouldn't timeout anyway.

2. This test doesn't construct an upstream reverse proxy! This set of tests just verifies the timeout handler behavior. To fix,  we could set that up here and add some more setup below or we need to move this to a set of tests where the the upstream reverse proxy is setup and used.

Jusshersmith: Great catch - I was a little bit too hasty in getting this added in!

PR #204, opened 2019-05-29
sso_auth: config name-spacing with viper [WIP]

Jusshersmith:
**[WIP]**

This is one possible solution for us to enable the name-spacing of configuration variables.

Effectively, this solution has us pass in a list of desired namespaces to `NewOptions()`, with separate structs for each being retuned in a map. (perhaps there can also be a way to simply return all configuration variables, in or not in a map of structs). From that map, relevant structs can be passed around where necessary.

(Specifics around actual namespaces used not yet solidified, the ones used were simply logical while implementing).

### DIFF
--- a/internal/auth/options.go
+++ b/internal/auth/options.go
@@ -124,16 +124,16 @@ type SignatureData struct {
 func NewOptions(prefix []string) (map[string]*Options, error) {
 	v := viper.New()
 
-	prefixMap := make(map[string]*Options)
+	opts := make(map[string]*Options)
 	for _, prefix := range prefix {
 		options, err := loadVars(v, prefix)
 		if err != nil {
 			return nil, err
 		}
-		prefixMap[prefix] = options
-
+		opts[prefix] = options
 	}
-	return prefixMap, nil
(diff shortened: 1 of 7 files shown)


internal/auth/options.go
```diff
@@ -205,46 +214,50 @@ func parseURL(toParse string, urltype string, msgs []string) (*url.URL, []string
 	return parsed, msgs
 }
 
+// REVIEW: I wonder if we would like to separate out these tests into namespace specific test functions?
+//         Preferred to keep one test function for all variables, or branch into separate provider/authenticator tests?
+
```
Jusshersmith: ^

jphines:
The overall namespacing seems fine, but I'm interested if we can move away from the underlying map structure to actual structs.

Jusshersmith:
Interesting. One other method I played around with was similar to this, but instead returning nested structs. I've just been testing separating out the tests/validations a little bit into separate namespaces - I'll see what I can do in regards to the underlying 'structure' the options are contained in!

PR #206, opened 2019-05-31
sso_auth: add default provider slug

jphines:
## Problem

After some reflection, we should add configuration for a default identity provider with `sso_auth` to make migrations easier and more seamless.

## Solution

Add additional configuration for the default provider slug and set up a default identity provider at the path root.

This also fixes `/robots.txt` which is mistakenly left dangling on the idp slug path.


internal/auth/mux.go
```diff
@@ -60,6 +60,25 @@ func NewAuthenticatorMux(opts *Options, statsdClient *statsd.Client) (*Authentic
 			fmt.Sprintf("/%s/", idpSlug),
 			http.StripPrefix(fmt.Sprintf("/%s", idpSlug), authenticator.ServeMux),
 		)
+
```
shrayolacrayon: can we add something to indicate that this is temporary and that we want to take this out eventually?

jphines: Added!

PR #207, opened 2019-06-03
authenticator: remove misc/cruft configuration

jphines:
## Problem
There is a ton of non-useful, unusable, or crufty configuration leftover in sso auth. This is an attempt to clean up the cruft configs.

### DIFF
--- a/internal/auth/options.go
+++ b/internal/auth/options.go
@@ -269,13 +269,6 @@ func newProvider(o *Options) (providers.Provider, error) {
 		ClientSecret:       o.ClientSecret,
 		ApprovalPrompt:     o.ApprovalPrompt,
 		SessionLifetimeTTL: o.SessionLifetimeTTL,
-
-		// set defaults so those aren't nil for future configuration
-		SignInURL:   &url.URL{},
-		RedeemURL:   &url.URL{},
-		RevokeURL:   &url.URL{},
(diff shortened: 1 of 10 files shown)


internal/auth/options.go
```diff
@@ -100,21 +85,11 @@ type Options struct {
...
 	// These options allow for other providers besides Google, with potential overrides.
 	Provider         string `mapstructure:"provider"`
 	ProviderSlug     string `mapstructure:"provider_slug"`
 	ProviderServerID string `mapstructure:"provider_server_id"`
 
-	SignInURL      string `mapstructure:"signin_url"`
-	RedeemURL      string `mapstructure:"redeem_url"`
-	RevokeURL      string `mapstructure:"revoke_url"`
-	ProfileURL     string `mapstructure:"profile_url"`
-	ValidateURL    string `mapstructure:"validate_url"`
```
Jusshersmith: With these removed, do we still want to provide the functionality for custom URL's to be passed in to providers? e.g: https://github.com/buzzfeed/sso/blob/master/internal/auth/providers/google.go#L45

jphines: That is my intent, yes -- I don't think this configuration is very useful if we also have to provide it with a provider implementation.

Jusshersmith: Great, agreed!

PR #211, opened 2019-06-12
ci: integrate with Codecov for test coverage

Jusshersmith:
## Problem

We already try hard to make sure code has appropriate tests written, however we don't currently provide an easy way to monitor and track test coverage.

## Solution

Integrate the sso repo with a tool like [Codecov](https://codecov.io/), which helps us track and monitor coverage reports and make them a prominent part of pull requests and the contribution workflow.

## Notes

This currently uses the Codecov [bash uploader](https://docs.codecov.io/v4.3.0/docs/about-the-codecov-bash-uploader) to upload coverage reports, which is one of the recommended methods. Other methods include updating CircleCI to version 2.1 and making use of CircleCI [orbs](https://circleci.com/blog/making-code-coverage-easy-to-see-with-the-codecov-orb/).

Specific settings for codecov to be decided - this PR has been created with defaults.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,4 @@
-version: 2.1
-orbs:
-  codecov: codecov/codecov@1.0.2
+version: 2
 jobs:
   build:
     docker:
@@ -29,11 +27,13 @@ jobs:
           command: |
             make dist/sso-proxy
       - run:
-            name: push sso-dev image
-            command: |
-              if [[ -n $DOCKER_USER ]]; then
-                docker login -u $DOCKER_USER -p $DOCKER_PASS
-                make imagepush
-              fi
-      - codecov/upload:
-          file: ./scripts/coverage.out
+          name: push sso-dev image
+          command: |
+            if [[ -n $DOCKER_USER ]]; then
+              docker login -u $DOCKER_USER -p $DOCKER_PASS
+              make imagepush
+            fi
+      - run:
+          name: upload coverage file to codecov
+          command: |
+            bash <(curl -s https://codecov.io/bash)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
-version: 2
+version: 2.1
+orbs:
+  codecov: codecov/codecov@1.0.2
 jobs:
   build:
     docker:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -27,9 +27,13 @@ jobs:
...
+          name: push sso-dev image
+          command: |
+            if [[ -n $DOCKER_USER ]]; then
+              docker login -u $DOCKER_USER -p $DOCKER_PASS
+              make imagepush
+            fi
+      - run:
+          name: upload coverage file to codecov
+          command: |
+            bash <(curl -s https://codecov.io/bash)
```
Jusshersmith: If we stick to this method, would we prefer to cache this file in CircleCI, or host it locally?

.codecov/codecov.yml
```diff
@@ -0,0 +1,23 @@
...
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "header, diff"
+  behavior: default
+  require_changes: no
+
```
Jusshersmith: This is largely default, with the only changes so far being:
- removal of notifications for the time being
- lowered the 'range' (preferred range, among other settings to be decided)

.circleci/config.yml
Jusshersmith: If we stick to this method, would we prefer to cache this file in CircleCI, or host it locally?

.codecov/codecov.yml
jphines: :+1:

.circleci/config.yml
jphines: This seems like the suggested method, right? let's just do this for now.  The next logical step would be to orb it.

jphines:
What do we need to merge this @Jusshersmith ?

Jusshersmith:
@jphines I think it's fine to merge as is - we'll work out the exact configuration we prefer as time goes on, but everything seems to be working well enough to start with! Also once this is merged I think the reports become more useful as they'll have master to compare against.

.codecov/codecov.yml
Jusshersmith: Also added a list of files that are impacted by the pull request and their coverage to the PR comment.

PR #213, opened 2019-06-13
[auth] allow proxying root domains

cameronattard:
## Problem

Currently, when users try to proxy requests to a root domain (i.e. not a subdomain of the root domain), they receive a 400 response (invalid redirect parameter). I would like to be able to serve content on my root domains in addition to the subdomains.

## Solution

Allow the literal root domain to be a valid redirect URI.

### DIFF
--- a/internal/auth/middleware.go
+++ b/internal/auth/middleware.go
@@ -129,7 +129,7 @@ func validRedirectURI(uri string, rootDomains []string) bool {
 		return false
 	}
 	for _, domain := range rootDomains {
-		if strings.HasSuffix(redirectURL.Hostname(), domain) {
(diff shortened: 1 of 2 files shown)


jphines:
@cameronattard Thanks for the contribution!

cameronattard:
My pleasure @jphines. Is there an ETA for the next release?

jphines:
@cameronattard we're planning on creating a formal release at the end of the month. You can pull dev-releases from https://hub.docker.com/r/buzzfeed/sso-dev -- but there is some drift now from latest docs in master to what the latest builds.

We're planning on cleaning those up as part of the formal release process.

PR #214, opened 2019-06-14
ci: remove minimum coverage setting

Jusshersmith:
## Problem

With the recent introduction of codecov, builds are failing to pass checks if their coverage falls below that of master.

## Solution

- Although we want to _increase_ coverage, I'm not sure if we want to set those hard limitations immediately?

    If I understand correctly these changes should remove the 'target', so we'll still get the PR comment but it won't fail due to coverage levels. 

- This also adds the codecov badge to our readme.

### DIFF
--- a/.codecov/codecov.yml
+++ b/.codecov/codecov.yml
@@ -1,11 +1,21 @@
+# Codecov configuration file for this repo.
+# This file can be validated for syntax and layout issues using:
+#     cat codecov.yml | curl --data-binary @- https://codecov.io/validate
+# as per Codecov's documentation.
+
+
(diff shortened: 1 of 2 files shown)


Jusshersmith:
Would we prefer to set a threshold perhaps? (allow the PR to drop by x% (5%?), but still pass)

jphines:
@Jusshersmith the threshold SGTM, maybe 2-3%?

Jusshersmith:
🙏 @jphines, please review/re-approve when you have the chance. Updated comments slightly to give some information around validating config.

PR #217, opened 2019-06-24
sso_auth: fix client authentication for revoke endpoint

Jusshersmith:
## Problem

`/oauth2/sign_out` does not currently work with the Okta provider:

> An error occurred during sign out. Please try again.

is returned, and the request to sign the user out is unsuccessful.

## Solution

This PR adds client authentication for the `Revoke` function (I think this must have been lost during development?), which is used to sign out via the `/oauth2/sign_out` endpoint.

## Notes

Client authentication is provided using the same method as in the `ValidateSessionState` function: https://github.com/buzzfeed/sso/blob/521ff9c1529f7015d9cdc4d2a75832192535db18/internal/auth/providers/okta.go#L117-L123

### DIFF
--- a/internal/auth/providers/okta.go
+++ b/internal/auth/providers/okta.go
@@ -413,20 +413,25 @@ func (p *OktaProvider) RefreshAccessToken(refreshToken string) (token string, ex
 	return
 }
 
-// Revoke revokes the access token a given session state.
+// Revoke revokes the refresh token from a given session state.
+// Revoking the refresh token implicitly revokes the access token, forcing re-authentication.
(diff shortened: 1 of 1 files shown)


Jusshersmith:
I believe we want to revoke the refresh token, which also forces the access token to be revoked:
> **Revoking only the access token** 
Revoking only the access token will effectively force the use of the refresh token to retrieve a new access token. This could be useful if, for example, you have changed a user's data and you want this information to be reflected in a new access token.

>**Revoking only the refresh token** 
If you revoke only the refresh token then the access token will also be revoked. This allows you to, for example, force a user to reauthenticate.

Issue #158, opened 2019-02-16
Support alternative cookie storage


sporkmonger:
Azure AD has massive tokens. It's obnoxious and unnecessary, but Microsoft isn't going to change this. Cookie compression helps, and cookie spanning solves most of the rest of the problem, but I still find that there are edge cases where it's a problem. For implementors that have these issues, I think it would be beneficial to just simply stop trying to do stateless session storage. If you can't reliably fit everything into 4kb, you really cannot guarantee that you won't have sign-in failures. Sometimes the failures are from browsers, sometimes it's from load balancers and other proxies enforcing their own limits. Regardless of where the limit is enforced, unfortunately, the user experience for a cookie size limit failure is bad.

I think SSO should support usage of a Redis backend. It ought to be pretty straightforward to point SSO to e.g. Elasticache Redis or an equivalent hosted option for most people who run into this issue, and that's going to be much more reliable than trying to jam the whole session directly into cookie state.

victornoel:
@sporkmonger see also #141 for a discussion on the matter :)

sylr:
Hi there,

Is anyone interested on working on this ? I'd do it myslef unfortunately I have no experience with Redis nor SSO knowledge.

Regards.

jphines:
Hi @sylr I don't believe anyone is working on this at the moment.

We'd be happy to accept and work through any contributions on this front!

brianv0:
This commit has an implementation of a Redis-backed cookie store I implemented in a fork of [pusher/oauth2_proxy](https://github.com/pusher/oauth2-proxy):

https://github.com/lsst-dm/oauth2_proxy/commit/29494561251efc7b1f066ef45455eb934630bcef

This implementation introduces a ServerCookiesStore interface:
```
type ServerCookiesStore interface {
	Store(responseCookie *http.Cookie, requestCookie *http.Cookie) (string, error)
	Clear(requestCookie *http.Cookie) error
	Load(requestCookie *http.Cookie) (string, error)
}
```

To which there's one implementation, a RedisCookiesStore.

This implementation introduces the concept of a Ticket, which is composed of three parts:

`{cookieName}-{uniqueId}.{iv}`

Where:
* `cookieName` is the same name configured to store the cookies in the browser.
* `uniqueId` is a random ID - it's found by just taking the sha1 of the cookie, but it could just be a unique ID/UUID or something
* `iv` is the initialization vector used to encrypt/decrypt the cookie in redis

[...]

PR #224, opened 2019-07-12
Redis sessions

sporkmonger:
## Problem

Session state can become quite large and can easily exceed the maximum cookie size. I previously tried to fix it with #150 but that proved to be a bad solution prone to failures.

## Solution

This introduces optional support for a Redis backend for sessions. Setting a `SESSION_REDIS_CONNECTION` environment variable to e.g. `redis://primary.sso-sessions.example-cloud.com:6379` will switch the session backend from cookie sessions to redis sessions. The default remains cookie sessions.

## Notes

This also avoids the issue of large amounts of session state being retransmitted on every request. Session state is encrypted using the existing cipher mechanism and session keys. As discussed in #158, I recommend renaming the cipher key configuration name to be more generic, but this PR doesn't make that change. This PR does not use the ticket system mentioned in #158, and instead opts for the simpler approach of encrypting using the session secret held on the auth server. I have tested this PR in conjunction with #118 and it works well.

### DIFF
--- a/internal/auth/configuration.go
+++ b/internal/auth/configuration.go
@@ -14,15 +14,18 @@ import (
 
 // DefaultAuthConfig specifies all the defaults used to configure sso-auth
 // All configuration can be set using environment variables. Below is a list of
-// configuration variables via their envivronment configuration
+// configuration variables via their environment configuration
 //
-// SESSION COOKIE_NAME
+// SESSION_COOKIE_NAME
 // SESSION_COOKIE_SECRET
 // SESSION_COOKIE_EXPIRE
 // SESSION_COOKIE_DOMAIN
 // SESSION_COOKIE_REFRESH
 // SESSION_COOKIE_SECURE
 // SESSION_COOKIE_HTTPONLY
+// SESSION_REDIS_CONNECTION
+// SESSION_REDIS_SENTINEL
+// SESSION_REDIS_MASTER
 // SESSION_LIFETIME
 // SESSION_KEY
 //
@@ -34,6 +37,9 @@ import (
 // PROVIDER_*_CLIENT_ID
 // PROVIDER_*_CLIENT_SECRET
 // PROVIDER_*_SCOPE
+//
+// PROVIDER_*_AZURE_TENANT
+// PROVIDER_*_AZURE_PROMPT
 //
 // PROVIDER_*_GOOGLE_CREDENTIALS
 // PROVIDER_*_GOOGLE_IMPERSONATE
@@ -83,6 +89,9 @@ func DefaultAuthConfig() Configuration {
 				Name:     "_sso_auth",
 				Secure:   true,
 				HTTPOnly: true,
+			},
+			RedisConfig: RedisConfig{
+				UseSentinel: false,
 			},
 		},
 		LoggingConfig: LoggingConfig{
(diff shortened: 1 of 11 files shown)


jphines:
@sporkmonger this is much cleaner than I expected, which is fantastic!

 I'll take a look at this today/this weekend, provide some review, and kick the tires.

Thank you for your work on this effort, it is a fantastic set of functionality. I (we) greatly appreciate your contributions!

sporkmonger:
Sure thing. I had some trouble figuring out how to mock the Redis client, which is why there's a bit of a coverage drop I believe.

internal/auth/configuration.go
```diff
@@ -35,6 +38,9 @@ import (
 // PROVIDER_*_CLIENT_SECRET
 // PROVIDER_*_SCOPE
 //
+// PROVIDER_*_AZURE_TENANT
+// PROVIDER_*_AZURE_PROMPT
```
sporkmonger: This snuck in somehow, not sure how critical it is to keep this isolated since this PR is the main blocker for Azure, but I can remove if needed.

sylr:
@sporkmonger it seems there currently is a conflict, could you resolve it ?

@jphines Do you have time to review this ?

jphines:
My apologizes, I've failed to prioritize review here appropriately. I've asked @Jusshersmith to review these changes in light of my inability to find the time to do so.

sporkmonger:
Yeah, I'll try to address the conflict shortly.

internal/proxy/proxy.go
```diff
@@ -43,9 +43,13 @@ func New(opts *Options) (*SSOProxy, error) {
 			return nil, err
 		}
 
+		setSessionStore := SetCookieStore
+		if opts.RedisConnectionURL != "" || opts.RedisUseSentinel {
+			setSessionStore = SetRedisStore
+		}
```
Jusshersmith: This is nice and concise, but I wonder if it's worth standardising to match the other similar use cases above, e.g:
```
if opts.RedisConnectionURL != "" || opts.RedisUseSentinel {
    optFuncs = append(optFuncs, SetRedisStore(opts)
} else {
    optFuncs = append(optFuncs, SetCookieStore(opts)
}
```
Or, at least closer to whatever method we end up using on the authenticator side to set the session store.

internal/auth/configuration_test.go
```diff
@@ -130,6 +130,57 @@ func TestEnvironmentOverridesConfiguration(t *testing.T) {
...
+		},
+		{
+			Name: "Test Redis Sessions",
+			EnvOverrides: map[string]string{
+				"SESSION_REDIS_CONNECTION": "redis://master.example.com:6379",
+			},
+			CheckFunc: func(c Configuration, t *testing.T) {
+				redisConf := c.SessionConfig.RedisConfig
+				assertEq([]string{"redis://master.example.com:6379"}, redisConf.ConnectionURLs, t)
+			},
```
Jusshersmith: I wonder if it's also worth adding in a call to RedisConfig.Validate() and/or RedisConfig.Enabled() here? It seems like it could be included fairly easily, and I think would help get the error cases within them tested.

Jusshersmith:
This looks great @sporkmonger, and thank you for your patience! I've left some comments to start with!

internal/auth/configuration.go
```diff
@@ -84,6 +90,9 @@ func DefaultAuthConfig() Configuration {
 				Secure:   true,
 				HTTPOnly: true,
 			},
+			RedisConfig: RedisConfig{
+				UseSentinel: false,
```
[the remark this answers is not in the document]
sporkmonger: Yes, that's correct. It's been awhile but I think it might fail if the `RedisConfig` struct doesn't get initialized, but the `UseSentinel` default is the `false` zero value.

sylr:
Any updates ?

sylr:
Guys ?

sporkmonger:
It's taken long enough that... we ended up moving from Azure AD to Okta in the meantime.

Jusshersmith:
Thank you for all of your great work on this @sporkmonger! This seems like it could still be an extremely valuable addition, and would be a shame for all of this work to go to waste. I don't think we at BuzzFeed necessarily have the capacity to push it over the finish line in the immediate future, but I think we can keep this open for a while still to allow someone to pick it up if possible -- providing you're happy with that @sporkmonger.

sporkmonger:
Totally. My bandwidth for stuff like this goes up and down a lot, but the next time I've got some spare cycles I'll try to get this addressed. If someone else wants to pick it up and get it over the finish line, I'm equally OK with that too.

PR #226, opened 2019-07-16
sso quickstart/docs: update configuration variables

Jusshersmith:
## Problem

We’ve made some substantial changes to the configuration variables (largely their names, but also what’s optional) required by SSO. This means that the quickstart and other documentation is outdated.

## Solution

This PR pulls our various documentation back in line with the configuration variables we now use. Most changes are applicable only to the ‘auth’ portion of SSO; the ‘proxy’ configuration variables have not yet been fully updated.

### DIFF
--- a/docs/google_provider_setup.md
+++ b/docs/google_provider_setup.md
@@ -43,8 +43,8 @@ asked to fill in three fields.
 - **Name**: Any appropriate name is fine (e.g. `Dev`).
 - **Authorized JavaScript origins**: Leave this field blank.
 - **Authorized redirect URIs**: Add the URI of your `sso-auth` deployment, with the path suffix
-`/oauth2/callback`. For example, if `sso-auth` will  be accessible at the domain
-`sso-auth.example.com`, then add the URI `https://sso-auth.example.com/oauth2/callback`.
+`/google/callback`. For example, if `sso-auth` will  be accessible at the domain
+`sso-auth.example.com`, then add the URI `https://sso-auth.example.com/google/callback`.
 
 **⚡️ Note:** If you're following the [Quickstart guide](quickstart.md), use
 `http://sso-auth.localtest.me` as the Authorized redirect URI.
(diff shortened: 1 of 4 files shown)


docs/google_provider_setup.md
```diff
@@ -122,7 +122,7 @@ have been filled in, click the "Authorize" button.
 To give `sso-auth` permission to access Google Group information for users, the following
 environment variables must be set:
 
-- **`GOOGLE_ADMIN_EMAIL`**: An administrative email address on your organization's
+- **`PROVIDER_GOOGLE_GOOGLE_IMPERSONATE`**: An administrative email address on your organization's
```
jphines: The first `GOOGLE` here doesn't have to be `GOOGLE` -- it can be any unique identifier that logical groups a set of provider configs together. This is confusing but maybe figuring out a way to call out?

docs/quickstart.md
```diff
@@ -33,14 +33,22 @@ section below:
...
+    PROVIDER_GOOGLE_CLIENT_ID=<client id value>.apps.googleusercontent.com
+    PROVIDER_GOOGLE_CLIENT_SECRET=<client secret value>
+
+As well as these, you also need to give a provider 'slug' for the configuration of the
+provider http routes.
+
+Add the below lines into the `env` file you just created:
+
+    PROVIDER_GOOGLE_TYPE=google
+    PROVIDER_GOOGLE_SLUG=google
```
jphines: Maybe we should make the identifier this slug instead of passing it in. Too magical?

docs/google_provider_setup.md
Jusshersmith: Yeah..Let me note that specifically and try to make that a bit clearer - maybe just using a different unique identifier for the quickstart will help.

Jusshersmith: @jphines - update the documentation to hopefully make this clearer

docs/quickstart.md
Jusshersmith: I like this idea. I don't personally feel like it's needed prior to the release (just in case anyone feels differently), but think it would be good to add!

PR #227, opened 2019-07-16
sso_auth: fix action tag parsing

Jusshersmith:
## Problem

Now that we include a Provider Slug in URL paths, the logic to parse and extract the ‘action’ from the URL incorrectly returns `unknown` for all (most) URL’s.

## Solution

Expecting a URL path of `/providerSlug/redeem` for example, split the path out based on `/` and test with the last slice item. This manually adds back the `/` (as the split removes it) in order to maintain a notion of URL paths within the overall logic.

## Notes

Our tests didn’t catch this - would be good to come up with a test that would have caught this.

### DIFF
--- a/internal/auth/metrics.go
+++ b/internal/auth/metrics.go
@@ -40,7 +40,10 @@ func GetActionTag(req *http.Request) string {
 	}
 	// get the action from the url path
 	path := req.URL.Path
-	if action, ok := pathToAction[path]; ok {
+	splitPath := strings.Split(path, "/")
+	pathBase := fmt.Sprintf("/%s", splitPath[len(splitPath)-1])
+
+	if action, ok := pathToAction[pathBase]; ok {
 		return action
 	}
 	if strings.HasPrefix(path, "/static/") {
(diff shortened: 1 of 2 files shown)


internal/auth/metrics.go
```diff
@@ -40,7 +40,8 @@ func GetActionTag(req *http.Request) string {
 	}
 	// get the action from the url path
 	path := req.URL.Path
-	if action, ok := pathToAction[path]; ok {
+	pathBase := strings.Split(path, "/")
+	if action, ok := pathToAction["/"+pathBase[len(pathBase)-1]]; ok {
```
jphines: can we do this on the previous line and pull out some of this logic. I also _strongly_ prefer `fmt.Sprintf` accurately

```go
splitPath := strings.Split(req.URL.PAth, "/")
path := fmt.Sprintf("/%s", splitPath[len(splitPath) - 1])
if action, ok := pathToAction[path]; ok {
. . .
}

Jusshersmith: That’s fair! 👍

PR #236, opened 2019-08-01
Relax the use of Email Domains/Addresses to enforce authentication

jphines:
## Problem

We currently mandate the use of a global email domain to authenticate both on the proxy side as well as the authenticator side. This is limiting for organizations and upstreams that have more diverse requirements that require a more flexible authenticator model.

We originally implemented this requirement as a safety precaution when SSO was less mature. It no longer provides the same assurances now that group usage is more robust and SSO itself has matured.

## Solution

We propose to move adjust this configuration in two ways:
1. Move the configuration in the proxy to the upstream configuration block.
2. Remove the configuration and mechanism on the proxy side. Instead, we will rely on the identity providers to provide this authentication mechanism.

## Notes

The way email domains, addresses, and groups interact with one another is becoming increasing confusing. We should think about ways to help simplify the model and make it more intuitive.

### DIFF
--- a/docs/sso_config.md
+++ b/docs/sso_config.md
@@ -30,7 +30,9 @@ For example, the following config would have the following environment variables
   * **to** is the cname of the proxied service (this tells sso proxy where to proxy requests that come in on the from field)
   * **type** declares the type of route to use, right now there is just *simple* and *rewrite*.
   * **options** are a set of options that can be added to your configuration.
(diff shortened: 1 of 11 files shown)


Jusshersmith:
This is largely just thinking out loud at the moment (I have no strong feelings), but is it worth allowing these configs to have a default set for them as well for where upstream specific settings are not required, but it's still wanted. (I suppose here: https://github.com/buzzfeed/sso/blob/55bf4486549af125ba616c42f6b2a3749962b66d/internal/proxy/options.go#L172).

I'm not entirely sure if that would be useful to many, or if it's just further solidifying something that _isn't_ all that useful anymore. (especially thinking of your comments in the 'Notes' section about simplifying this stuff)

Issue #243, opened 2019-08-12
Allow injection of request headers in proxy


mcfearsome:
**Is your feature request related to a problem? Please describe.**
I am setting sso up in front of a Kibana instance that is created using the ECK operator.  This does not allow me to turn off the xpack security feature (I can on Kibana but am unable to for Elasticsearch which causes issues for Kibana).  In order to avoid having the user login twice I would like to be able to include the basic auth credentials in the upstream config.

**Describe the solution you'd like**
Add new option to the `UpstreamConfig` similar to `HeaderOverrides`

**Describe alternatives you've considered**
I could not use the ECK operator and run ES without any security at all, would prefer to not do that

jphines:
@mcfearsome I believe setting `HeaderOverrides` will work in this case to inject new headers. Does it not?

mcfearsome:
@jphines `HeaderOverrides` looks to be for Response Headers not adding onto the request before it hits the proxied service. I have a PR incoming (never wrote Go before so go easy)

mcfearsome:
Just to make sure I wasn't being dumb I double checked using `HeaderOverrides` with the current value that is working with `InjectRequestHeaders` and it does not work (but is present in the Response Headers)

PR #244, opened 2019-08-12
proxy: Allow injection of request headers

mcfearsome:
* Added InjectRequestHeaders to UpstreamConfig/OptionsConfig
* Loop over InjectRequestHeaders during Authenticate and add headers to the request

## Problem
#243 

## Notes

I am unfamiliar with Go and was unsure how to go about testing the actual adding of the headers.  If someone would be kind enough to point me in the right direction I can continue.

### DIFF
--- a/internal/proxy/oauthproxy.go
+++ b/internal/proxy/oauthproxy.go
@@ -802,6 +802,9 @@ func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) (er
 	req.Header.Set("X-Forwarded-Email", session.Email)
 	req.Header.Set("X-Forwarded-Groups", strings.Join(session.Groups, ","))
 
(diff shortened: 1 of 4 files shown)


internal/proxy/oauthproxy.go
```diff
@@ -802,6 +802,9 @@ func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) (er
 	req.Header.Set("X-Forwarded-Email", session.Email)
 	req.Header.Set("X-Forwarded-Groups", strings.Join(session.Groups, ","))
 
+	for key, val := range p.upstreamConfig.InjectRequestHeaders {
```
jphines: Let's move these above setting of the `X-Forwarded-` headers. We don't want to allow impersonation by setting these headers from a config.

mcfearsome: moved

PR #246, opened 2019-08-20
sso_auth: add default for groupcache settings

Jusshersmith:
## Problem 

If `c.refreshTTL` is not greater than 0 then a panic is caused upon the execution of https://github.com/buzzfeed/sso/blob/9c169b34603a63883f26dc2b717e65df34b0503f/internal/pkg/groups/fillcache.go#L118

At the moment, there's no default for this so it must be explicitly set to avoid a panic.

## Solution

Set a sane default for `refreshTTL` (this default was removed during the overhaul of configuration options).


jphines:
:shipit:

PR #247, opened 2019-08-23
sso_proxy: ability to define allowed email address/domain in upstream config

Jusshersmith:
## Problem
Allowed email addresses and domains can only be specified for SSO as a whole, setting the same restriction on all upstreams. It cannot be set on individual upstreams.

## Solution

Allow these variables to be defined in two separate places:
- in the environment (in the standard 'options' config), acting as a default setting for all upstreams
- within individual upstream configs, overriding the 'default' setting

We still expect one of these to be set and will error if neither are.

## Notes

Includes small bug fix to both the email address validator and email domain validator packages to avoid mutating the passed in slice, instead opting to create and append to a new one.

### DIFF
--- a/internal/proxy/options.go
+++ b/internal/proxy/options.go
@@ -24,8 +24,8 @@ import (
 // Cluster - the cluster in which this is running, used for upstream configs
 // Scheme - the default scheme, used for upstream configs
 // SkipAuthPreflight - will skip authentication for OPTIONS requests, default false
-// EmailDomains - csv list of emails with the specified domain to authenticate. Use * to authenticate any email
-// EmailAddresses - []string - authenticate emails with the specified email address (may be given multiple times). Use * to authenticate any email
+// DefaultAllowedEmailDomains - csv list of emails with the specified domain to authenticate. Use * to authenticate any email
+// DefaultAllowedEmailAddresses - []string - authenticate emails with the specified email address (may be given multiple times). Use * to authenticate any email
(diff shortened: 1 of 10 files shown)


internal/proxy/oauthproxy.go
```diff
@@ -193,6 +194,13 @@ func NewOAuthProxy(opts *Options, optFuncs ...func(*OAuthProxy) error) (*OAuthPr
 			return nil, err
 		}
 	}
+	// if allowed EmailDomains or EmailAddresses are defined in the upstream config,
+	// override any previously set emailValidators.
+	if len(p.upstreamConfig.EmailAddresses) != 0 {
+		p.emailValidator = options.NewEmailAddressValidator(p.upstreamConfig.EmailAddresses)
+	} else if len(p.upstreamConfig.EmailDomains) != 0 {
+		p.emailValidator = options.NewEmailDomainValidator(p.upstreamConfig.EmailDomains)
+	}
```
jphines: Would should supply a defensive closing else statement to make sure we're always passing in a validator.

internal/proxy/options.go
```diff
@@ -147,8 +147,9 @@ func (o *Options) Validate() error {
 	if o.ClientSecret == "" {
 		msgs = append(msgs, "missing setting: client-secret")
 	}
+	var noDomainOrAddressOverride bool
```
jphines: This seems like a code smell. Let's specify these as a required default values and allow overrides per upstream.

PR #250, opened 2019-08-29
auth: amazon cognito provider

jphines:
This adds an additional Identity Provider for Amazon Cognito.

There are some TODO's here:

* Sign out behavior for Amazon Cognito is a bit weird -- there is no endpoint to issues sign-outs except globally.
* There is no validation endpoint -- we make the full profile call to validate user state
* Group behavior is a big TODO. It would be nice if we could make this a custom attribute, but something we should look into
* Documentation


internal/auth/providers/amazon_cognito.go
```diff
@@ -0,0 +1,395 @@
...
+		ExpiresIn    int64  `json:"expires_in"`
+		IDToken      string `json:"id_token"`
+	}
+
+	err := p.amazonCognitoRequest("POST", p.RedeemURL.String(), params, []string{"action:redeem"}, nil, true, &response)
+	if err != nil {
+		return nil, err
+	}
+
+	logger.Info(fmt.Sprintf("response %#v", response))
```
Jusshersmith: Do we want to always print out the response here?

internal/auth/providers/amazon_cognito.go
```diff
@@ -0,0 +1,395 @@
...
+func (p *AmazonCognitoProvider) RefreshSessionIfNeeded(s *sessions.SessionState) (bool, error) {
+	if s == nil || !s.RefreshPeriodExpired() || s.RefreshToken == "" {
+		return false, nil
+	}
+
+	newToken, duration, err := p.RefreshAccessToken(s.RefreshToken)
+	if err != nil {
+		return false, err
+	}
+	logger := log.NewLogEntry()
```
Jusshersmith: nit: just to keep this standardised with other functions, move this logger line to right below the function signature I guess?

Jusshersmith: it does also appear to be in this spot within the other providers too ¯\_(ツ)_/¯

internal/auth/providers/amazon_cognito.go
```diff
@@ -0,0 +1,395 @@
...
+	return userinfo.EmailAddress, nil
+}
+
+// ValidateGroupMembership takes in an email, a list of allowed groups an access token
+// - compares allowed groups to users' groups.
+// Conditionally returns an empty (nil) slice or a slice of the matching groups.
+func (p *AmazonCognitoProvider) ValidateGroupMembership(email string, allowedGroups []string, accessToken string) ([]string, error) {
+	// TODO: This is currently non-functioning
+	return allowedGroups, nil
+}
```
Jusshersmith: I'm not sure what we'd want the behaviour to be in the case of no group membership functionality. At the moment this effectively means the group validation check always passes, right?

internal/auth/providers/amazon_cognito.go
```diff
@@ -0,0 +1,395 @@
...
+// For the Amazon Cognito provider this triggers a global sign out for this user on all devices
+func (p *AmazonCognitoProvider) Revoke(s *sessions.SessionState) error {
+	_, err := p.cip.GlobalSignOut(&cognitoidentityprovider.GlobalSignOutInput{
+		AccessToken: &s.AccessToken,
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
Jusshersmith: Because this doesn't need to use the `amazonCognitoRequest` function, we lose the metric tags that are normally passed in with it. Is it worth adding those tags back in manually here?

internal/auth/providers/amazon_cognito_admin.go
```diff
@@ -0,0 +1,270 @@
...
+	if err != nil {
+		logger.WithError(err).Fatal("error creating aws session")
+		return nil, err
+	}
+	idp := cognitoidentityprovider.New(sess)
+
+	return idp, nil
+}
+
+//TODO: we could potentially remove this in favour of the GetUserInfo cognito provider method
```
Jusshersmith: Just to highlight this `TODO`. We could remove it, though it's also quite nice having the admin module contain everything needed?

go.mod
```diff
@@ -14,9 +15,10 @@ require (
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/rakyll/statik v0.1.6
 	github.com/sirupsen/logrus v1.4.2
+	golang.org/x/net v0.0.0-20190620200207-3b0461eec859 // indirect
 	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
 	golang.org/x/sync v0.0.0-20190423024810-112230192c58
-	golang.org/x/xerrors v0.0.0-20190513163551-3ee3066db522
+	golang.org/x/xerrors v0.0.0-20190717185122-a985d3407aa7
```
Jusshersmith: These changes can be removed now

Jusshersmith:
👋 @jphines - pinging you on this as although you wrote the first portion, a fair chunk (and time) has been added since then, so seems worth another general look over, providing you have the bandwidth?

I think the biggest things that have been added/changed is the group membership validation logic (https://github.com/buzzfeed/sso/pull/250/files?file-filters%5B%5D=.go#diff-ea62f132a9d50a2ecbeb99d1dd8c48e1R340), 
and the cognito admin file (https://github.com/buzzfeed/sso/pull/250/files?file-filters%5B%5D=.go#diff-c8e9125622d7201b686c3d99338c8e3aR1). Along with some related configuration additions.

internal/auth/providers/amazon_cognito_admin.go
```diff
@@ -0,0 +1,270 @@
...
+type CognitoAdminService struct {
+	adminService *cognitoidentityprovider.CognitoIdentityProvider
+	StatsdClient *statsd.Client
+	cb           *circuit.Breaker
+	userPoolID   *string
+}
+
+func getCognitoIdentityProvider(id, secret, region string) (*cognitoidentityprovider.CognitoIdentityProvider, error) {
+	logger := log.NewLogEntry()
+	sess, err := session.NewSession(&aws.Config{
```
jphines: since we have our _own_ package named `session` -- it's probably worth adding an alias so it is explicit this is an aws session pkg. Through myself there for a loop first time I read this 😨

internal/auth/providers/amazon_cognito_admin.go
jphines: Looks like we implemented this by calling the admin service anyway -- so it seems reasonable to just use that and we can delete this func.

internal/auth/providers/amazon_cognito_admin.go
```diff
@@ -0,0 +1,270 @@
...
+
+	if accessToken == "" {
+		return nil, ErrBadRequest
+	}
+	reqParams := &cognitoidentityprovider.GetUserInput{
+		AccessToken: &accessToken,
+	}
+
+	startTS := time.Now()
+	req, userinfo := cas.adminService.GetUserRequest(reqParams)
```
jphines: I think good "golang" style would caml case `userInfo`

internal/auth/providers/amazon_cognito_admin.go
```diff
@@ -0,0 +1,270 @@
...
+			break
+		}
+		nextToken = *resp.NextToken
+	}
+	return groupMembers, nil
+
+}
+
+func (cas *CognitoAdminService) CheckMemberships(userName string) ([]string, error) {
+
```
jphines: ✂️ this newline

internal/auth/providers/amazon_cognito_admin.go
Jusshersmith: Oh yeah...hah. Bet that was confusing at first 😬

internal/auth/providers/amazon_cognito_admin.go
```diff
@@ -0,0 +1,217 @@
...
+	return inGroups, nil
+}
+
+func (cas *CognitoAdminService) GlobalSignOut(session *sessions.SessionState) error {
+	tags := []string{
+		"provider:cognito",
+		"action:global_sign_out",
+	}
+
+	req, _ := cas.adminService.GlobalSignOutRequest(&cognitoidentityprovider.GlobalSignOutInput{
```
jphines: Should we not check the error here?

internal/auth/providers/amazon_cognito.go
```diff
@@ -0,0 +1,472 @@
...
+// AmazonCognitoProvider is an implementation of the Provider interface.
+type AmazonCognitoProvider struct {
+	*ProviderData
+	StatsdClient *statsd.Client
+	AdminService CognitoAdminProvider
+	cb           *circuit.Breaker
+	GroupsCache  groups.MemberSetCache
+}
+
+type GetCognitoUserProfileResponse struct {
```
jphines: Since we only use this struct in this pkg lets not export it by making it lowercase. `getCognitoUserProfileResponse`

internal/auth/providers/amazon_cognito.go
jphines: meh lets make it right below the func sig

internal/auth/providers/amazon_cognito.go
```diff
@@ -0,0 +1,472 @@
...
+	if err != nil {
+		return false
+	}
+
+	return true
+}
+
+// GetSignInURL returns the sign in url with typical oauth parameters
+func (p *AmazonCognitoProvider) GetSignInURL(redirectURI, state string) string {
+	// https://developer.okta.com/docs/api/resources/oidc/#authorize
```
jphines: wrong docs link

internal/auth/providers/amazon_cognito.go
```diff
@@ -0,0 +1,472 @@
...
+		Path:   "/oauth2/authorize",
+	}
+	// returns access tokens, ID tokens or refresh tokens
+	p.RedeemURL = &url.URL{
+		Scheme: scheme,
+		Host:   OrgURL,
+		Path:   "/oauth2/token",
+	}
+
+	// TODO: Cognito doesn't provide a revoke endpoint to log out users
```
jphines: Do we have to do this by the global signout?

jphines: yah -- let's write a note here that this is how revocation is done

jphines:
A few comment nits -- but the logic looks good.

internal/auth/providers/amazon_cognito_admin.go
Jusshersmith: As far as I understand it, because we're using the `Send` method on the request (`req`), it's the `Send` call output that holds the error. The output that is ignored above contains the _response_ if `Send` returns without error. 

Brief explanation in the docs here: https://docs.aws.amazon.com/sdk-for-go/api/service/cognitoidentityprovider/#CognitoIdentityProvider.GlobalSignOutRequest

> The "output" return value will be populated with the request's response once the request completes successfully.
> Use "Send" method on the returned Request to send the API call to the service. the "output" return value is not valid until after Send returns without error.

In this case though, the sign out response should be empty: https://docs.aws.amazon.com/sdk-for-go/api/service/cognitoidentityprovider/#GlobalSignOutOutput

jphines: Oh, this isn't an error! Its a response! Weird API.

PR #253, opened 2019-09-23
sso_*: allow simultaneous use of Validators

Jusshersmith:
## Problem

The connection between `AllowedEmailDomains`, `AllowedEmailAddresses`, and `AllowedGroups` is unclear. If all specified, they don't work together well which causes awkward workarounds to be put into place.

## Solution

Introduce a more structured 'Validator' type within SSO, allowing us to create better and clearer dynamics between these settings (and any other validators that may be added in the future).

There are currently three separate validators: 
- Email Domain
- Email Address
- Email Group

**SSO will now allow the request through providing at least _one_ of these validators pass.**

## Notes

- I would suggest looking at the following files together: 
```
- internal/pkg/options/validators.go
- internal/pkg/options/email_address_validator.go
- internal/pkg/options/email_domain_validator.go
- internal/pkg/options/email_group_validator.go
```
```
- internal/proxy/proxy.go
- internal/proxy/oauthproxy.go
```
```
- internal/auth/authenticator.go
- internal/auth/mux.go
- internal/auth/options.go
```
```
- internal/pkg/options/mock_validator.go
- internal/auth/authenticator_test.go
- internal/proxy/oauthproxy_test.go
```
```
- internal/pkg/options/email_address_validator_test.go
- internal/pkg/options/email_domain_validator_test.go
```

- This also changes the use of the `*` wildcard within `AllowedEmailDomains` and `AllowedEmailAddresses`. **The wildcard will be ignored if other domains/email addresses are also passed with it, choosing to follow the most restrictive path.**

- On second thoughts..I think it may have been clearer to recreate the validator files from scratch!

### DIFF
--- a/internal/auth/authenticator_test.go
+++ b/internal/auth/authenticator_test.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/buzzfeed/sso/internal/auth/providers"
 	"github.com/buzzfeed/sso/internal/pkg/aead"
+	"github.com/buzzfeed/sso/internal/pkg/options"
 	"github.com/buzzfeed/sso/internal/pkg/sessions"
 	"github.com/buzzfeed/sso/internal/pkg/templates"
 	"github.com/buzzfeed/sso/internal/pkg/testutil"
@@ -62,13 +63,6 @@ func setMockAuthCodeCipher(cipher *aead.MockCipher, s interface{}) func(*Authent
 func setTestProvider(provider *providers.TestProvider) func(*Authenticator) error {
 	return func(a *Authenticator) error {
 		a.provider = provider
-		return nil
-	}
-}
-
-func setMockValidator(response bool) func(*Authenticator) error {
-	return func(a *Authenticator) error {
-		a.Validator = func(string) bool { return response }
 		return nil
 	}
 }
@@ -424,7 +418,7 @@ func TestSignIn(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			config := testConfiguration(t)
 			auth, err := NewAuthenticator(config,
-				setMockValidator(tc.validEmail),
+				SetValidators([]options.Validator{options.NewMockValidator(tc.validEmail)}),
 				setMockSessionStore(tc.mockSessionStore),
 				setMockTempl(),
 				setMockRedirectURL(),
@@ -571,7 +565,7 @@ func TestSignOutPage(t *testing.T) {
 			provider.RevokeError = tc.RevokeError
 
 			p, _ := NewAuthenticator(config,
-				setMockValidator(true),
+				SetValidators([]options.Validator{options.NewMockValidator(true)}),
 				setMockSessionStore(tc.mockSessionStore),
 				setMockTempl(),
 				setTestProvider(provider),
@@ -948,7 +942,7 @@ func TestGetProfile(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			config := testConfiguration(t)
 			p, _ := NewAuthenticator(config,
-				setMockValidator(true),
+				SetValidators([]options.Validator{options.NewMockValidator(true)}),
 			)
 			u, _ := url.Parse("http://example.com")
 			testProvider := providers.NewTestProvider(u)
@@ -1050,7 +1044,7 @@ func TestRedeemCode(t *testing.T) {
 			config := testConfiguration(t)
 
 			proxy, _ := NewAuthenticator(config,
-				setMockValidator(true),
+				SetValidators([]options.Validator{options.NewMockValidator(true)}),
 			)
 
 			testURL, err := url.Parse("example.com")
@@ -1357,7 +1351,8 @@ func TestOAuthCallback(t *testing.T) {
 					Value: "state",
 				},
 			},
-			expectedError: HTTPError{Code: http.StatusForbidden, Message: "Invalid Account"},
+			expectedError: HTTPError{Code: http.StatusForbidden,
+				Message: "We ran into some issues while validating your account: \"MockValidator error\""},
 		},
 		{
 			name: "valid email, invalid redirect",
@@ -1438,7 +1433,7 @@ func TestOAuthCallback(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			config := testConfiguration(t)
 			proxy, _ := NewAuthenticator(config,
-				setMockValidator(tc.validEmail),
+				SetValidators([]options.Validator{options.NewMockValidator(tc.validEmail)}),
 				setMockCSRFStore(tc.csrfResp),
 				setMockSessionStore(tc.sessionStore),
 			)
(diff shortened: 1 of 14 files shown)


internal/pkg/options/validators.go
```diff
@@ -0,0 +1,33 @@
...
+	ErrInvalidEmailAddress = errors.New("Invalid Email Address In Session State")
+	ErrValidationError     = errors.New("Error during validation")
+)
+
+type Validator interface {
+	Validate(*sessions.SessionState) error
+	Type() string
+}
+
+func RunValidators(validators []Validator, session *sessions.SessionState) ([]string, []error) {
```
jphines: We can clean up this function definition by just returning errors. It doesn't look like returning the validator type as a string slice buys us much?

internal/pkg/options/validators.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+var (
+	ErrInvalidEmailAddress = errors.New("Invalid Email Address In Session State")
+	ErrValidationError     = errors.New("Error during validation")
+)
+
+type Validator interface {
+	Validate(*sessions.SessionState) error
+	Type() string
```
jphines: This `Type() string` is a code smell. If we want to infer the type of something we should be using type checks rather than passing around a string.

internal/pkg/options/validators.go
Jusshersmith: The main thing it provides is logging around which validators passed (without having the validators emit log messages themselves). It felt like it might be something useful to log now that we allow the request as long as one validator is successful. 

I'm happy to remove though -- perhaps the benefit isn't worth it in this case!

internal/pkg/options/validators.go
Jusshersmith: Removed this (check the previous comment for extra information)

internal/pkg/options/validators.go
Jusshersmith: ^ removed!

internal/proxy/oauthproxy.go
```diff
@@ -567,40 +565,23 @@ func (p *OAuthProxy) OAuthCallback(rw http.ResponseWriter, req *http.Request) {
...
+
+	errors := options.RunValidators(p.Validators, session)
+	if len(errors) != 0 {
+		tags = append(tags, "error:validation_failed")
 		p.StatsdClient.Incr("application_error", tags, 1.0)
 		logger.WithRemoteAddress(remoteAddr).WithUser(session.Email).Info(
-			"permission denied: unauthorized")
-		p.ErrorPage(rw, req, http.StatusForbidden, "Permission Denied", "Invalid Account")
+			fmt.Sprintf("permission denied: unauthorized: %q", errors))
+		// TODO: As is, the error page given to users provides no extra contract around what failed.
```
Jusshersmith: 1. *context
2. This seems like an opportunity to refine the error pages that running these validators will show, and make sure they match where wanted. (depending on where in the logic these validators were ran we _might_ want different errors shown to the end user?) At the moment, there's a couple different variations.

internal/auth/authenticator_test.go
```diff
@@ -424,7 +418,7 @@ func TestSignIn(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			config := testConfiguration(t)
 			auth, err := NewAuthenticator(config,
-				setMockValidator(tc.validEmail),
+				SetValidators([]options.Validator{options.NewMockValidator(tc.validEmail)}),
```
Jusshersmith: Would we prefer to stick with a custom 'setValidator' function or similar to use within these tests?

internal/proxy/oauthproxy_test.go
```diff
@@ -1336,3 +1321,78 @@ func TestHTTPSRedirect(t *testing.T) {
 		})
 	}
 }
+
```
Jusshersmith: Need to check this out

internal/proxy/oauthproxy_test.go
```diff
@@ -1336,3 +1321,78 @@ func TestHTTPSRedirect(t *testing.T) {
...
+// via the upstream config will override those in the environment.
+//func TestAllowedEmailAddressOverrides(t *testing.T) {
+//	testCases := []struct {
+//		name                      string
+//		environmentEmailAddresses string
+//		upstreamEmailAddresses    []string
+//		expectedStatusCode        int
+//	}{
+//		{
+//			name:                      "environmentEmailAddress set, upstreamEmailAddress unset",
```
Jusshersmith: Need to check this test out

internal/proxy/oauthproxy.go
```diff
@@ -567,40 +565,23 @@ func (p *OAuthProxy) OAuthCallback(rw http.ResponseWriter, req *http.Request) {
...
+		p.ErrorPage(rw, req, http.StatusForbidden, "Permission Denied", "Account Validation Error")
 		return
```
Jusshersmith: 1. *context
2. This seems like an opportunity to refine the error pages that running these validators will show, and make sure they match where wanted. (depending on where in the logic these validators were ran we might want different errors shown to the end user?) At the moment, there's a couple different variations.

Jusshersmith:
Some additional test files are also needed, i.e: `internal/pkg/options/email_group_validator_test.go`

internal/auth/authenticator_test.go
jphines: This seems fine as written.

internal/proxy/oauthproxy.go
jphines: It would be nice if the error page surfaced this information!

internal/proxy/oauthproxy_test.go
jphines: This test verified override behavior from the default configuration to the resolved state, ensuring that the defaults were set correctly and overridden by upstream specific configuration.

internal/proxy/oauthproxy.go
```diff
@@ -567,40 +565,29 @@ func (p *OAuthProxy) OAuthCallback(rw http.ResponseWriter, req *http.Request) {
...
-		logger.WithRemoteAddress(remoteAddr).WithUser(session.Email).WithAllowedGroups(
-			allowedGroups).Info("permission denied: unauthorized")
-		p.ErrorPage(rw, req, http.StatusForbidden, "Permission Denied", "Group membership required")
+		formattedErrors := make([]string, 0, len(errors))
+		for _, err := range errors {
+			formattedErrors = append(formattedErrors, err.Error())
+		}
+		errorMsg := fmt.Sprintf("We ran into some issues while validating your account: \"%s\"",
+			strings.Join(formattedErrors, ", "))
+		p.ErrorPage(rw, req, http.StatusForbidden, "Permission Denied", errorMsg)
```
Jusshersmith: TODO: Not a huge issue, but if any of the errors are duplicated (which is currently possible) then they'll also be duplicated on the error page -- which we could easily avoid

internal/proxy/oauthproxy_test.go
Jusshersmith: This was added and accidentally pulled over from a separate branch -- it's a useful test to have, but not very relevant to this PR. I've removed it and will work on testing + re-adding it separately.

internal/proxy/oauthproxy.go
jphines: SGTM. We could clean up the visual a little bit but let's do that in a follow up.

PR #261, opened 2019-10-23
sso_*: move sign_in and sign_out HTML pages to sso_proxy

Jusshersmith:
## Problem
(PR description and contents incomplete)
The current setup of html sign in and sign out pages causes some extra friction with implementing https://github.com/buzzfeed/sso/pull/252 - this change gives us some extra flexibility. We can also simplify the flow a little between the proxy <-> authenticator by moving these pages to the proxy. 

## Solution

Moves the sign_in and sign_out HTML pages to be rendered by sso_proxy instead of sso_auth. 

## Notes

There are still some specific flows and edge cases that need to be considered within this change, as well as tests that need to be updated/added.

We'll also soon look to improve the usefulness of these pages (among other html pages), and the sign in/out process in a separate PR.


Jusshersmith:
Marked as ready for review, but see Notes in description.

PR #262, opened 2019-10-23
cmd: ensure http servers shut down gracefully

mccutchen:
## Problem

Both `sso-proxy` and `sso-auth` abort all in-flight requests on shutdown.  Instead, they should stop accepting new requests and allow in-flight requests to complete before exiting.

Addressing this will make deploys of the sso services less disruptive to end users.

## Solution

Because graceful shutdown is _just_ tricky enough to get right, here we add a new internal `httpserver` package that exposes a single function, `Run()`, which will run an http server and arrange for graceful shutdown when receiving `SIGTERM` or `SIGINT`.  A shutdown timeout must also be provided, after which the server will abort any outstanding requests.

### Reproducing

The easiest way I know to reproduce the behavior uses httpbin via the quickstart:
 1. In the quickstart dir, `docker-compose up -d` to bring up the demo services
 2. In a browser, visit http://httpbin.sso.localtest.me/delay/9 to start a request that will take 9 seconds to complete
 3. Before those 9 seconds elapse, stop the sso-proxy container (note the `-t 30` to tell docker to wait up to 30 seconds before forcibly killing it):
   ```
   docker stop -t 30 $(docker ps | grep /bin/sso-proxy | awk '{print $1}')
   ```

On latest master, you should immediately get an HTTP 502 error.

With these changes, you should instead get a successful response after the 9 seconds have elapsed, and _then_ sso-proxy will be shut down.

### Unit tests

I'm not entirely sure how to test this, given that
1. it relies on the process receiving either `SIGTERM` or `SIGINT`, and
2. the simple `httpserver.Run()` interface manages the whole lifecycle of the server, leaving no straightforward way to control the listener or figure out what port it is listening on (assuming we'd want it to choose a random available port)

If anybody has any good ideas for tackling (1) I'm all ears; the issues in (2) are under our control, so we could conceivably make the interface more flexible and use a trick [like this](https://stackoverflow.com/a/48283226) to choose a random port in advance (though this would be inherently race-y).

I'll also leave a few notes inline below.

### DIFF
--- a/cmd/sso-auth/main.go
+++ b/cmd/sso-auth/main.go
@@ -6,15 +6,16 @@ import (
 	"os"
 
 	"github.com/buzzfeed/sso/internal/auth"
-	log "github.com/buzzfeed/sso/internal/pkg/logging"
+	"github.com/buzzfeed/sso/internal/pkg/httpserver"
+	"github.com/buzzfeed/sso/internal/pkg/logging"
 )
 
 func init() {
-	log.SetServiceName("sso-authenticator")
+	logging.SetServiceName("sso-authenticator")
 }
 
 func main() {
-	logger := log.NewLogEntry()
+	logger := logging.NewLogEntry()
 
 	config, err := auth.LoadConfig()
 	if err != nil {
@@ -53,5 +54,7 @@ func main() {
 		Handler:      auth.NewLoggingHandler(os.Stdout, timeoutHandler, config.LoggingConfig.Enable, statsdClient),
 	}
 
-	logger.Fatal(s.ListenAndServe())
+	if err := httpserver.Run(s, config.ServerConfig.TimeoutConfig.Shutdown, logger); err != nil {
+		logger.WithError(err).Fatal("error running server")
+	}
 }
--- a/cmd/sso-proxy/main.go
+++ b/cmd/sso-proxy/main.go
@@ -8,17 +8,18 @@ import (
 
 	"github.com/kelseyhightower/envconfig"
 
-	log "github.com/buzzfeed/sso/internal/pkg/logging"
+	"github.com/buzzfeed/sso/internal/pkg/httpserver"
+	"github.com/buzzfeed/sso/internal/pkg/logging"
 	"github.com/buzzfeed/sso/internal/proxy"
 	"github.com/buzzfeed/sso/internal/proxy/collector"
 )
 
 func init() {
-	log.SetServiceName("sso-proxy")
+	logging.SetServiceName("sso-proxy")
 }
 
 func main() {
-	logger := log.NewLogEntry()
+	logger := logging.NewLogEntry()
 
 	opts := proxy.NewOptions()
 	err := envconfig.Process("", opts)
@@ -58,5 +59,7 @@ func main() {
 		Handler:      loggingHandler,
 	}
 
-	logger.Fatal(s.ListenAndServe())
+	if err := httpserver.Run(s, opts.ShutdownTimeout, logger); err != nil {
+		logger.WithError(err).Fatal("error running server")
+	}
 }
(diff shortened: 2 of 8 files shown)


cmd/sso-auth/main.go
```diff
@@ -42,9 +44,12 @@ func main() {
 	}
 	defer authMux.Stop()
 
+	requestTimeout := config.ServerConfig.TimeoutConfig.Request
+	shutdownTimeout := requestTimeout + 1*time.Second
```
mccutchen: I'm not sure there's any real value in having the shutdown timeout be a little longer than the request timeout we already apply to this server.  I don't think there's much harm in this, but 1 second is pretty arbitrary and I'd be happy to drop this and just use the request timeout as the shutdown timeout.

cmd/sso-proxy/main.go
```diff
@@ -58,5 +59,10 @@ func main() {
 		Handler:      loggingHandler,
 	}
 
-	logger.Fatal(s.ListenAndServe())
+	// How long we will wait to allow in-flight requests to finish
+	shutdownTimeout := 30 * time.Second
```
mccutchen: For sso-auth above, there's a natural shutdown timeout to use based on the request timeout we already configure at the server level.  For sso-proxy, I'm not sure there's any existing upper-limit on request time that we can use here, so I picked a fairly arbitrary number.

(30 seconds matches the time that we at BuzzFeed have configured the AWS ecs agent to wait before forcibly killing a container when it does not exit after receiving `SIGTERM`.)

We could also make this an explicit config value w/ a reasonable default, if you'd prefer.  Or, please let me know if there is some existing config value on which it would make sense to piggyback here!

go.sum
```diff
@@ -97,6 +97,7 @@ github.com/google/pprof v0.0.0-20190515194954-54271f7e092f/go.mod h1:zfwlbNMJ+OI
 github.com/google/renameio v0.1.0/go.mod h1:KWCgfxg9yswjAJkECMjeO8J8rahYeXnNhOm40UhjYkI=
 github.com/google/uuid v1.0.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/google/uuid v1.1.1/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
+github.com/googleapis/gax-go/v2 v2.0.4 h1:hU4mGcQI4DaAYW+IbTun+2qEZVFxK0ySjQLTbS0VQKc=
```
mccutchen: 🤷‍♂ my local golang toolchain kept wanting to add this, so I stopped fighting it

internal/pkg/httpserver/httpserver.go
```diff
@@ -0,0 +1,47 @@
...
+
+		close(exitCh)
+	}()
+
+	if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		return err
+	}
+
+	<-exitCh
+	logger.Info("shutdown finished")
```
mccutchen: I have found log messages like these handy when using logs to try to put together timelines of service behavior, but they're not all that helpful in general use.  If this seems too noisy, we can drop the logging from this function altogether.

internal/pkg/httpserver/httpserver.go
```diff
@@ -0,0 +1,47 @@
...
+	exitCh := make(chan struct{})
+
+	go func() {
+		sig := <-shutdownCh
+		logger.Info("shutdown started by signal: ", sig)
+
+		ctx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+		defer cancel()
+
+		srv.SetKeepAlivesEnabled(false)
```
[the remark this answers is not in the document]
mccutchen: I can't!  I _thought_ I had copied this from [the stdlib's Shutdown example](https://golang.org/pkg/net/http/#Server.Shutdown), but looking at it again I do not see this called.

Went ahead and dropped this.

mccutchen:
@jphines thanks for the feedback, I think I've addressed all your comments.

> I'm not entirely sure how to test this

So I did add some tests, please let me know what you think of the approach I took.  In short, in the unit tests we redefine the set of shutdown signals that `httpserver.Run()` listens for, and send the new shutdown signal to the test process to exercise this behavior.

Note, though, the giant dumb caveat here, explaining why this doesn't perfectly test the behavior, and why I'm not sure the extra complexity to perfectly test it is worth it:
https://github.com/buzzfeed/sso/blob/1bf00ac759096617fa55c130254dede17e3c7eef/internal/pkg/httpserver/httpserver_test.go#L49-L76

internal/pkg/httpserver/httpserver.go
jphines: sgtm

jphines:
@mccutchen I agree with that the additional complexity necessary to "properly" test the signal handling doesn't seem worth it at this point in time.

This seems reasonable to land as-is.

Thank you so much for the contribution, this is a great changeset. 💯 🚀

PR #266, opened 2019-11-07
validator pkg: control when each validator is ran

Jusshersmith:
## Problem

With the validator abstraction work that was recently done we inadvertently started to run group validations more than we used to.

Depending on the request volume flowing through SSO running the group check _again_ can cause issues with upstream providers. 

## Solution

We don't need to validate the groups again here. This pull request adds a feature to the validator package that allows us to pass in a flag whenever calling the `RunValidators()` function that will determine which validators we want to run in that flow. 

This gives us more flexibility to control when specific validators should or shouldn't be ran.

## Notes

Now that the group membership check is an official 'validator' within sso-proxy it's ran each time we call `RunValidators()`, whereas before when running the equivalent function the group check wasn't part of this.

Specifically, here: https://github.com/buzzfeed/sso/blob/9019d4f79b453b50882213baa8549d40852daaf7/internal/proxy/oauthproxy.go#L784 

Previously, we were only checking email address/domains as the group check is ran just above that when refreshing or validating the session: https://github.com/buzzfeed/sso/blob/9019d4f79b453b50882213baa8549d40852daaf7/internal/proxy/oauthproxy.go#L731 & https://github.com/buzzfeed/sso/blob/9019d4f79b453b50882213baa8549d40852daaf7/internal/proxy/oauthproxy.go#L762

### DIFF
--- a/internal/auth/authenticator.go
+++ b/internal/auth/authenticator.go
@@ -226,7 +226,7 @@ func (p *Authenticator) authenticate(rw http.ResponseWriter, req *http.Request) 
 		}
 	}
 
-	errors := options.RunValidators(p.Validators, session)
+	errors := options.RunValidators(p.Validators, options.AuthenticatorFlow, session)
 	if len(errors) == len(p.Validators) {
 		logger.WithUser(session.Email).Info(
 			fmt.Sprintf("permission denied: unauthorized: %q", errors))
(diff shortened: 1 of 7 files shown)


Jusshersmith:
Closing in favour of https://github.com/buzzfeed/sso/pull/267

PR #271, opened 2019-11-21
fillcache pkg: trigger cache update immediately

Jusshersmith:
## Problem
The fillcache package is what we use with the Google (and soon to be Cognito) providers to keep group membership cached. 

Once a request goes through to `sso_auth` to validate a users group membership, the `ValidateGroup` function looks in the local cache for the relevant groups:
https://github.com/buzzfeed/sso/blob/14f2a962d995bbd4b69da4420df5fe53dc3468c2/internal/auth/providers/google.go#L366-L370 
If the group doesn't exist in the cache, it triggers a refresh of the cache which in turn kicks off a goroutine to keep the group cache up to date (based off an input TTL).

Right now, the triggered goroutine only fills the cache _after_ the given TTL, not immediately. Any subsequent calls to `ValidateGroupMembership` between the first call, and the first tick of `ttl` won’t have the groups cached.
## Solution

Break out the logic that the goroutine uses to fill the cache into a separate method, and call it once _before_ the loop, and then upon each tick of `ttl`. 

Going to look at updating this PR with some logging examples to better verify the bug + expected behaviour 

## Notes
- There are few different ways we could implement this, but the last commit includes one that is perhaps most _clear_

### DIFF
--- a/internal/pkg/groups/fillcache.go
+++ b/internal/pkg/groups/fillcache.go
@@ -19,7 +19,7 @@ type MemberSetCache interface {
 	// Get returns a MemberSet from the cache
 	Get(string) (MemberSet, bool)
 	// Update updates the MemberSet of a given key, return a boolean updated value, and and error
-	Update(string) (bool, error)
+	Update(string) bool
 	// RefreshLoop starts an update refresh loop for a given key and returns a boolean value of it was started
 	RefreshLoop(string) bool
 	// Stop is a function to stop all goroutines that may have been spun up for the cache.
@@ -71,11 +71,14 @@ func (c *FillCache) Get(group string) (MemberSet, bool) {
 // Update recomputes the value for the given key, unless another goroutine is
 // already computing the value, and returns a bool indicating whether the value
 // was updated
-func (c *FillCache) Update(group string) (bool, error) {
+func (c *FillCache) Update(group string) bool {
+	logger := log.NewLogEntry()
+	logger.WithUserGroup(group).Info("updating fill cache")
+
 	c.mu.Lock()
 	if _, waiting := c.inflight[group]; waiting {
 		c.mu.Unlock()
-		return false, nil
+		return false
 	}
 
 	c.inflight[group] = struct{}{}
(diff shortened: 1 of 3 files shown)


internal/pkg/groups/fillcache.go
```diff
@@ -93,6 +93,27 @@ func (c *FillCache) Update(group string) (bool, error) {
 	return false, err
 }
 
+// update wraps the Update method, adding instrumentation
+func (c *FillCache) update(group string) {
```
Jusshersmith: Very open to any thoughts around making this new method 'fit in' a little bit more

jphines: we say this "wraps" the `Update` method -- but I don't see any wrapping here?

jphines: 🤔 Would it be better to just add this instrumentation to the `Update` method?

Jusshersmith: I did think that, but the original setup felt pretty nice to me - to have the instrumentation separated. Happy to move it into the `Update` method if you think that'd be cleaner still.

Jusshersmith: Looking again, maybe it's not so bad? I'll give it a quick try and see how it fits.

Jusshersmith: Added some potential changes to bring some of the instrumentation into the `Update` method -- it now only returns a `bool`, instead of a `bool` and `error`. The error seemed to only be used _for_ the instrumentation, though it was being tested against in some capacity (purely when testing the functionality, not in 'production'. 

Maybe there's a reason we want to keep returning those errors though

internal/pkg/groups/fillcache.go
```diff
@@ -71,11 +71,14 @@ func (c *FillCache) Get(group string) (MemberSet, bool) {
...
 // already computing the value, and returns a bool indicating whether the value
 // was updated
-func (c *FillCache) Update(group string) (bool, error) {
+func (c *FillCache) Update(group string) bool {
+	logger := log.NewLogEntry()
+	logger.WithUserGroup(group).Info("updating fill cache")
+
 	c.mu.Lock()
 	if _, waiting := c.inflight[group]; waiting {
 		c.mu.Unlock()
```
Jusshersmith: maybe we should add a log line here for more observability into it.

internal/pkg/groups/fillcache.go
```diff
@@ -19,7 +19,7 @@ type MemberSetCache interface {
 	// Get returns a MemberSet from the cache
 	Get(string) (MemberSet, bool)
 	// Update updates the MemberSet of a given key, return a boolean updated value, and and error
-	Update(string) (bool, error)
+	Update(string) bool
```
jphines: If we remove the error, lets update the comment above `, and and error`

Jusshersmith:
Now that https://github.com/buzzfeed/sso/pull/273 is merged, we're back and once again ✅!

PR #273, opened 2019-12-02
sso: update go modules

Jusshersmith:
## Problem
Updates to certain go modules are required. This is starting to cause errors with new builds: https://circleci.com/gh/buzzfeed/sso/2185?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

## Solution

Running https://github.com/buzzfeed/sso/blob/master/scripts/test locally edited `go.mod` and `go.sum` files as required.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,8 @@ ENV GO111MODULE=on
 WORKDIR /go/src/github.com/buzzfeed/sso
 
 COPY . .
-RUN cd cmd/sso-auth && go build --mod=readonly -o /bin/sso-auth
-RUN cd cmd/sso-proxy && go build --mod=readonly -o /bin/sso-proxy
+RUN cd cmd/sso-auth && go build -mod=readonly -o /bin/sso-auth
+RUN cd cmd/sso-proxy && go build -mod=readonly -o /bin/sso-proxy
 
 # =============================================================================
 # final stage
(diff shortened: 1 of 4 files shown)


mccutchen:
Should we also add `-mod=readonly` to the `go build` commands in our Dockerfile and Makefile to more explicitly require up-to-date module information?

Jusshersmith:
Yeah, I think that sounds like a sensible idea @mccutchen!

Dockerfile
```diff
@@ -12,8 +12,8 @@ ENV GO111MODULE=on
 WORKDIR /go/src/github.com/buzzfeed/sso
 
 COPY . .
-RUN cd cmd/sso-auth && go build -o /bin/sso-auth
-RUN cd cmd/sso-proxy && go build -o /bin/sso-proxy
+RUN cd cmd/sso-auth && go build --mod=readonly -o /bin/sso-auth
+RUN cd cmd/sso-proxy && go build --mod=readonly -o /bin/sso-proxy
```
mccutchen: Nit, here and below:  using the single-dash form is more idiomatic for go commands: `-mod` instead of `--mod`.

Jusshersmith: Oh interesting! Thanks for pointing out!

Jusshersmith:
Think I'm going to need a fresh review on this @mccutchen! (thanks!)

PR #274, opened 2019-12-05
sso_proxy: fix request signer hash panic

Jusshersmith:
## Problem

Occasional panics (`http: panic serving 172.17.0.16:49478: d.nx != 0`) which seem to be caused by `sso_proxy` using the same hasher for multiple requests, creating a race condition between the `hasher.Reset()` and `hasher.Write([]byte(repr))` calls.

## Solution

Instead of storing a `hash.Hash` object within the `RequestSigner`, instead store a function which creates a new `hash.Hash`. This lets each request be signed using a different Hash.

### DIFF
--- a/internal/proxy/request_signer.go
+++ b/internal/proxy/request_signer.go
@@ -38,7 +38,7 @@ var signingKeyHeader = "kid"
 // RequestSigner exposes an interface for digitally signing requests using an RSA private key.
 // See comments for the Sign() method below, for more on how this signature is constructed.
 type RequestSigner struct {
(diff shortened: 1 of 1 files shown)


jphines:
🤦‍♂

Jusshersmith:
Rebased and removed `go.mod` and `go.sum` changes, now that https://github.com/buzzfeed/sso/pull/273 is merged.

Jusshersmith:
I think I've said the word 'hash' a few too many times now 🤔

PR #275, opened 2019-12-06
sso_proxy: reduce direct calls to ValidateGroup() and clean up logic

Jusshersmith:
## Problem

We are still calling `ValidateGroup()` directly within `sso_proxy`, but using the options/validator package elsewhere in the same logic path (originally partially due to circular imports). This makes it increasingly difficult to make sure we were running the right validations at the right time, and certain methods were growing in complexity and responsibility.

## Solution
Attempt to reunite some of the most problematic portions of code in related to the above.

High level overview of included changes:
- Moves `extendDeadline` and `withinGracePeriod` to be part of the sessions package, instead of the providers package. 
- Changes direct calls to `ValidateGroup` within `internal/proxy/providers/sso.go` to options/validator package calls within `internal/proxy/providers/oauthproxy.go`.
   - Introduces the `runValidatorsWithGracePeriod` helper method here to help handle cases where we want to check if the auth provider is unavailable instead of explicitly denying authentication.
- Renames the `ValidateSessionState` method to `ValidateSessionToken`, which seemed to better fit its responsibility. 

## Notes
The perhaps less obvious change is that within the `Authenticate()` method we'll now only run  validators when the refresh or validation period has expired. 
This is instead of running group validations when the refresh or validation period has expired, and domain/email validations on all proxied requests.


internal/proxy/oauthproxy.go
```diff
@@ -781,25 +829,6 @@ func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) (er
 		}
 	}
 
-	// We revalidate group membership whenever the session is refreshed or revalidated
```
jphines: I'm on board with not having to run these validators on every request. However, there is a part of me that's anxious about removing these checks.

What validation do we do to authenticate and authorize this request to this upstream beyond these checks? I want to make sure that at least it's impossible to take a valid token from upstream foo and use it against upstream bar.

internal/proxy/oauthproxy.go
```diff
@@ -375,6 +375,27 @@ func (p *OAuthProxy) IsWhitelistedRequest(req *http.Request) bool {
 	return false
 }
 
+// runValidatorsWithGracePeriod runs all validators and upon finding errors, checks to see if the
+// auth provider is explicity denying authentication or if it's merely unavailable. If it's unavailable,
+// we check whether the session is within the grace period or not to determine the specific error we return.
+func (p *OAuthProxy) runValidatorsWithGracePeriod(session *sessions.SessionState) (err error) {
+	logger := log.NewLogEntry()
+	errors := options.RunValidators(p.Validators, session)
+	if len(errors) == len(p.Validators) {
```
jphines: This logic seems wrong to me. If every validator fails, we want to return nil and allow the user?

internal/proxy/oauthproxy.go
```diff
@@ -375,6 +375,27 @@ func (p *OAuthProxy) IsWhitelistedRequest(req *http.Request) bool {
 	return false
 }
 
+// runValidatorsWithGracePeriod runs all validators and upon finding errors, checks to see if the
+// auth provider is explicity denying authentication or if it's merely unavailable. If it's unavailable,
+// we check whether the session is within the grace period or not to determine the specific error we return.
+func (p *OAuthProxy) runValidatorsWithGracePeriod(session *sessions.SessionState) (err error) {
```
jphines: It seems like we only every run validators with the grace period in the proxy. It also seems strange to me that the function that is responsible for running the validators hangs of the options struct, which doesn't feel like it has much to do with running the validators?

Can we move the running the validators to the OAuthProxy struct and have it just always run the grace period? I think this simplification will better delineate the responsibility of these various structs and simplifying the confusing naming.

internal/proxy/oauthproxy.go
Jusshersmith: I think (unless _I'm_ misunderstanding/misreading) the `return nil` on line `396` is only called if _not_ all validators failed. Too many `{ }`'s everywhere 😬. Maybe a simple newline there will make that a little bit clearer.

Is that what you were looking at?

internal/proxy/oauthproxy.go
Jusshersmith: On the point of `options` in particular, I think in this case that is the 'options' _package_, the naming of which hung around from previous times. The reason that we're using the package there specifically is to call the `RunValidators` helper function, with the validators that are stored in the OAuthProxy struct being passed in.


I think now that these are more defined 'validators' the package could surely do with a name update.

internal/proxy/providers/providers.go
```diff
@@ -13,10 +13,10 @@ type Provider interface {
 	Redeem(string, string) (*sessions.SessionState, error)
 	ValidateGroup(string, []string, string) ([]string, bool, error)
 	UserGroups(string, []string, string) ([]string, error)
-	ValidateSessionState(*sessions.SessionState, []string) bool
+	ValidateSessionToken(*sessions.SessionState) bool
 	GetSignInURL(redirectURL *url.URL, finalRedirect string) *url.URL
 	GetSignOutURL(redirectURL *url.URL) *url.URL
-	RefreshSession(*sessions.SessionState, []string) (bool, error)
+	RefreshSession(*sessions.SessionState) (bool, error)
```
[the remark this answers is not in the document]
Jusshersmith: 👍 Yeah - seems sensible!

internal/proxy/oauthproxy.go
jphines: I'm sorry -- its been a few weeks. Is the logic here OR logic, if any of the validators pass, we want to allow the user?

internal/proxy/oauthproxy.go
jphines: ah, right -- I think the point still holds regardless of the package vs struct.

internal/proxy/oauthproxy.go
Jusshersmith: No worries! That's correct - allow the user if any of the defined validators pass.

internal/proxy/oauthproxy.go
Jusshersmith: I've renamed the 'options' package to 'validators' to better represent its responsibility, though given that change I'm not sure if we'd still like to change the how we run the validators themselves?

Right now we're running these validators in two slightly different ways within the proxy:
- by calling the `validators.RunValidators` function in the `OAuthCallback` method (https://github.com/buzzfeed/sso/pull/275/files#diff-0f5a3bf10e8922fb438df82f694747a5L572)
- by calling the `(oauthproxy.)runValidatorsWithGracePeriod` method in the `Authenticate` method, which calls the `validators.RunValidators` function before adding some extra logic to handle provider grace periods
(https://github.com/buzzfeed/sso/pull/275/files#diff-0f5a3bf10e8922fb438df82f694747a5R781)

internal/pkg/sessions/session_state.go
```diff
@@ -45,6 +46,14 @@ func (s *SessionState) ValidationPeriodExpired() bool {
 	return isExpired(s.ValidDeadline)
 }
 
+// IsWithinGracePeriod returns true if the session is still within the grace period
+func (s *SessionState) IsWithinGracePeriod(gracePeriodTTL time.Duration) bool {
```
jphines: This is a good refactoring. This PR points out how we could in a followup could/should convert this so that this Period acts like our other periods. 💯

internal/proxy/oauthproxy.go
```diff
@@ -375,6 +375,27 @@ func (p *OAuthProxy) IsWhitelistedRequest(req *http.Request) bool {
...
+	logger := log.NewLogEntry()
+	errors := validators.RunValidators(p.Validators, session)
+	if len(errors) == len(p.Validators) {
+		for _, err := range errors {
+			// Check to see if the auth provider is explicity denying authentication, or if it is merely unavailable.
+			if err == providers.ErrAuthProviderUnavailable && session.IsWithinGracePeriod(p.provider.Data().GracePeriodTTL) {
+				return err
+			}
+		}
+		allowedGroups := p.upstreamConfig.AllowedGroups
```
jphines: Can we put the allowed groups in the validator error message instead of pulling them out this way? I think it could make the logline easier to understand which set of validators rejected a user?

internal/proxy/oauthproxy.go
```diff
@@ -589,6 +610,9 @@ func (p *OAuthProxy) OAuthCallback(rw http.ResponseWriter, req *http.Request) {
 	logger.WithRemoteAddress(remoteAddr).WithUser(session.Email).WithInGroups(session.Groups).Info(
 		fmt.Sprintf("oauth callback: user validated "))
 
+	//TODO: is there any verification we want to do on this?
```
jphines: I think this is sufficient -- let's drop the TODO and a comment about why we add this new value to session

internal/proxy/oauthproxy.go
```diff
@@ -757,9 +803,11 @@ func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) (er
...
 		// Validation period has expired, this is the shortest interval we use to
 		// check for valid requests. This should be set to something like a minute.
-		// This calls up the provider chain to validate this user is still active
-		// and hasn't been de-authorized.
-		ok := p.provider.ValidateSessionState(session, allowedGroups)
+		// In this case we:
+		// - call up the provider chain to validate this user is still active and hasn't been de-authorized.
+		// - run any defined email domain, email address, and email group validators against the session
+
+		ok := p.provider.ValidateSessionToken(session)
```
jphines: This refactoring also makes me wonder if this function signature should be the same as refresh? Presumably the validation of the session token could fail in the same two ways:

1. We just are physically unable to validate the session token (i.e. it error'd)
2. We successfully validated the session but the user is no longer authorized (bool is false)

Let's make a note that we can clean this up in a following PR? Overall a good set of refactors!

internal/proxy/oauthproxy.go
Jusshersmith: Added the allowed groups into the group validator error message. Below is an example of the error message displayed to users (the same formatting is used for the log lines).

Screen Shot 2020-01-28 at 12 11 50

Also removed a chunk of formatting logic around the errors as it was becoming over-engineered and unnecessary. We could also add some extra context to errors coming from the domain/email address validators, but as is the error message would become pretty bloated if multiple validators returned errors - so perhaps worth addressing that separately?

Jusshersmith:
@jphines - ready for re-review. Main new logic changes consist of the change to the raised error if an unauthorised upstream is being requested to better handle the graceful introduction of this check, and formatting of the errors returned by the validators.

Other than that, it's largely comment changes/additions.

internal/proxy/oauthproxy.go
```diff
@@ -653,6 +676,12 @@ func (p *OAuthProxy) Proxy(rw http.ResponseWriter, req *http.Request) {
 			// the user has a stale sesssion.
 			p.OAuthStart(rw, req, tags)
 			return
+		case ErrUnauthorizedUpstreamRequested:
+			// The users session has been authorised for use with a different upstream than the one
+			// that is being requested, so we trigger the start of the oauth flow.
+			// This exists primarily to implement some form of grace period while this additional session
+			// check is being introduced.
+			p.OAuthStart(rw, req, tags)
```
jphines: We need a return statement here after this:

```golang
p.OAuthStart(rw, req, tags)
return
```

internal/proxy/oauthproxy.go
```diff
@@ -757,9 +808,13 @@ func (p *OAuthProxy) Authenticate(rw http.ResponseWriter, req *http.Request) (er
...
 		// Validation period has expired, this is the shortest interval we use to
 		// check for valid requests. This should be set to something like a minute.
-		// This calls up the provider chain to validate this user is still active
-		// and hasn't been de-authorized.
-		ok := p.provider.ValidateSessionState(session, allowedGroups)
+		// In this case we:
+		// - call up the provider chain to validate this user is still active and hasn't been de-authorized.
+		// - run any defined email domain, email address, and email group validators against the session
+
+		//TODO: change this to match the RefreshSessionToken method
```
jphines: This looks done now?

internal/proxy/oauthproxy.go
Jusshersmith: Good catch.

internal/proxy/oauthproxy.go
Jusshersmith: Slightly misleading comment unless the commented link is followed - I'll update for clarity

PR #212, opened 2019-06-12
sso: support multiple identity providers

jphines:
## Problem

We've been walking down the path to support multiple identity providers in several previous iterations. This last remaining stand is how to support _configuration_ for multiple and different identity providers.  The existing configuration structure is pretty inflexible and rigid when it comes to defining configuration, especially given our constraint to use environment variables.

We previously introduced `viper` to help solve some of these problems, but viper proved frustrating and lacking enough opinion to operate within. We roll that solution back here and instead use https://micro.mu/docs/go-config.html

## Solution

We introduced a new configuration component based on https://micro.mu/docs/go-config.html to build this new, more complicated configuration mechanism.  This is a ground-up re-write of the configuration mechanism we use, found separately in `configuration.go`.

## Notes

This now looks like:
```
### session
SESSION COOKIE_NAME
SESSION_COOKIE_SECRET
SESSION_COOKIE_EXPIRE
SESSION_COOKIE_DOMAIN
SESSION_COOKIE_REFRESH
SESSION_COOKIE_SECURE
SESSION_COOKIE_HTTPONLY
SESSION_LIFETIME
SESSION_KEY


### client
CLIENT_PROXY_ID
CLIENT_PROXY_SECRET


### provider config for google
PROVIDER_*_TYPE
PROVIDER_*_SLUG
PROVIDER_*_CLIENT_ID
PROVIDER_*_CLIENT_SECRET
PROVIDER_*_SCOPE

### google specific 
PROVIDER_*_GOOGLE_CREDENTIALS
PROVIDER_*_GOOGLE_IMPERSONATE

### okta specific
PROVIDER_*_OKTA_URL
PROVIDER_*_OKTA_SERVER

### group refresh
PROVIDER_*_GROUPCACHE_INTERVAL_REFRESH
PROVIDER_*_GROUPCACHE_INTERVAL_PROVIDER


# server config
SERVER_SCHEME
SERVER_HOST
SERVER_PORT
SERVER_TIMEOUT_REQUEST
SERVER_TIMEOUT_WRITE
SERVER_TIMEOUT_READ


# authorize config
AUTHORIZE_PROXY_DOMAINS
AUTHORIZE_EMAIL_DOMAINS
AUTHORIZE_EMAIL_ADDRESSES


# metrics config 
METRICS_STATSD_PORT
METRICS_STATSD_HOST


# logging config
LOGGING_ENABLE
LOGGING_LEVEL
```

PR #279, opened 2020-01-13
sso_proxy: update to use `go-micro` for configuration management

Jusshersmith:
## Problem

`sso_auth` was migrated to use `go-micro` in https://github.com/buzzfeed/sso/pull/212 to manage variable configuration. `sso_proxy` is still using the original outdated methods and should be updated to also use `go-micro`

## Solution

Update `sso_proxy` to use `go-micro`. No feature changes are intended to be part of this change, *apart from* the behaviour of the following two variables:
- `SKIP_AUTH_PREFLIGHT` and `PASS_ACCESS_TOKEN` are now **upstream specific** and should be specified as part of upstream configs, **not** global/environment variables.

In general, the patterns used in this pull request are the same ones used in https://github.com/buzzfeed/sso/pull/212

## Review notes
Some notes to support reviewing:
- The bulk of changes here consist of creating the new configuration structure, and altering functions and methods to reference that new structure.
- Numerous functions have been moved out of `oauthproxy.go` and into `options.go`:
  - `SetCookieStore`
  - `SetRequestSigner`
  - `SetUpstreamConfig`
  - `SetProxyHandler`
  - `SetValidators`
  - `SetProvider`

- A new function, `SetUpstreamConfigs`, largely consists of some logic that was in the `options.Validate` method.  

#### The below table shows all old variables, and their new equivalent version.

|   old	| new  	|
|---	|---	|
| **session configs** | 
|  COOKIE_NAME 	 |   SESSION_COOKIE_NAME	|
|   COOKIE_SECRET  | SESSION_COOKIE_SECRET   	|
|   COOKIE_EXPIRE 	 |   SESSION_COOKIE_EXPIRE	|
|   COOKIE_DOMAIN	|   SESSION_COOKIE_DOMAIN 	|
|   COOKIE_HTTP_ONLY	 |   SESSION_COOKIE_HTTPONLY	|
| COOKIE_SECURE | SESSION_COOKIE_SECURE |
|   SESSION_LIFETIME_TTL   |   SESSION_TTL_LIFETIME|
|   SESSION_VALID_TTL   |   SESSION_TTL_VALID	|
|   GRACE_PERIOD_TTL   |      SESSION_TTL_GRACEPERIOD|
| --- | 
| **upstream configs**	|
|   DEFAULT_ALLOWED_EMAIL_DOMAINS	|   UPSTREAM_DEFAULT_EMAIL_DOMAINS	|
|  DEFAULT_ALLOWED_EMAIL_ADDRESSES 	|   UPSTREAM_DEFAULT_EMAIL_ADDRESSES	|
|   DEFAULT_ALLOWED_GROUPS	|   UPSTREAM_DEFAULT_GROUPS	|
|   DEFAULT_UPSTREAM_TIMEOUT	|   UPSTREAM_DEFAULT_TIMEOUT	|
|   DEFAULT_UPSTREAM_TCP_RESET_DEADLINE	|   UPSTREAM_DEFAULT_RESETDEADLINE	|
|   	|   	|
|   UPSTREAM_CONFIGS	|   UPSTREAM_CONFIGSFILE	|
| | |
|   CLUSTER	|   UPSTREAM_CLUSTER	|
| SCHEME | UPSTREAM_SCHEME |
| PROVIDER | UPSTREAM_DEFAULT_PROVIDER |
| --- |
|   **server configs**	|
|   PORT	|  SERVER_PORT 	|
|   SHUTDOWN_TIMEOUT	|   SERVER_TIMEOUT_SHUTDOWN	|
|   TCP_READ_TIMEOUT	|   SERVER_TIMEOUT_READ	|
|   TCP_WRITE_TIMEOUT	|   SERVER_TIMEOUT_WRITE	|
| --- |
| **metrics configs**	|
|   STATSD_PORT	|   METRICS_STATSD_HOST	|
|   STATSD_HOST	|   METRICS_STATSD_PORT	|
| --- |
| **logging configs** |
|   REQUEST_LOGGING	|   LOGGING_ENABLE	|
|   	|   ~LOGGING_LEVEL~	|
| --- |
| **client configs**	|
|   CLIENT_ID	|   CLIENT_ID	|
|   CLIENT_SECRET	|   CLIENT_SECRET	|
| --- |
|**request signature configs**	|
|   REQUEST_SIGNATURE_KEY	|   REQUESTSIGNER_KEY	|
| --- |
| **provider configs**|
|   PROVIDER	|   PROVIDER_TYPE	|
|   PROVIDER_URL	|   PROVIDER_URL_EXTERNAL	|
|   PROVIDER_URL_INTERNAL | PROVIDER_URL_INTERNAL |
|   SCOPE   | PROVIDER_SCOPE  |

### DIFF
--- a/internal/proxy/configuration.go
+++ b/internal/proxy/configuration.go
@@ -1,0 +1,460 @@
+package proxy
+
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/micro/go-micro/config"
+	"github.com/micro/go-micro/config/source/env"
+	"github.com/mitchellh/mapstructure"
+)
+
+// DefaultProxyCongig specifies all the defaults used to configure sso-proxy
+// All configuration can be set using environment variables. Below is a list of
+// configuration variables via their environment configuraiton
+//
+// SESSION_COOKIE_NAME
+// SESSION_COOKIE_SECRET
+// SESSION_COOKIE_EXPIRE
+// SESSION_COOKIE_DOMAIN
+// SESSION_COOKIE_HTTPONLY
+// SESSION_TTL_LIFETIME
+// SESSION_TTL_VALID
+// SESSION_TTL_GRACEPERIOD
+//
+// REQUESTSIGNER_KEY
+//
+// CLIENT_ID
+// CLIENT_SECRET
+//
+// SERVER_PORT
+// SERVER_TIMEOUT_SHUTDOWN
+// SERVER_TIMEOUT_READ
+// SERVER_TIMEOUT_WRITE
+//
+// METRICS_STATSD_HOST
+// METRICS_STATSD_PORT
+//
+// LOGGING_ENABLE
+//
+// UPSTREAM_DEFAULT_EMAIL_DOMAINS
+// UPSTREAM_DEFAULT_EMAIL_ADDRESSES
+// UPSTREAM_DEFAULT_GROUPS
+// UPSTREAM_DEFAULT_TIMEOUT
+// UPSTREAM_DEFAULT_TCP_RESET_DEADLINE
+// UPSTREAM_DEFAULT_PROVIDER
+// UPSTREAM_CONFIGS_FILE
+// UPSTREAM_SCHEME
+// UPSTREAM_CLUSTER
+//
+// PROVIDER_TYPE
+// PROVIDER_URL_EXTERNAL
+// PROVIDER_URL_INTERNAL
+// PROVIDER_SCOPE
+
+func DefaultProxyConfig() Configuration {
+	return Configuration{
+		ServerConfig: ServerConfig{
+			Port: 4180,
+			TimeoutConfig: TimeoutConfig{
+				Write:    30 * time.Second,
+				Read:     30 * time.Second,
+				Shutdown: 30 * time.Second,
+			},
+		},
+		ProviderConfig: ProviderConfig{
+			ProviderType: "sso",
+		},
+		SessionConfig: SessionConfig{
+			CookieConfig: CookieConfig{
+				Name:     "_sso_proxy",
+				Expire:   168 * time.Hour,
+				Secure:   true,
+				HTTPOnly: true,
+			},
+			TTLConfig: TTLConfig{
+				Lifetime:    720 * time.Hour,
+				Valid:       60 * time.Second,
+				GracePeriod: 3 * time.Hour,
+			},
+		},
+		UpstreamConfigs: UpstreamConfigs{
+			DefaultConfig: DefaultConfig{
+				Timeout:       10 * time.Second,
+				ResetDeadline: 60 * time.Second,
+				ProviderSlug:  "google",
+			},
+			Scheme: "https",
+		},
+		LoggingConfig: LoggingConfig{
+			Enable: true,
+		},
+		MetricsConfig: MetricsConfig{
+			StatsdConfig: StatsdConfig{
+				Port: 8125,
+				Host: "localhost",
+			},
+		},
+	}
+}
+
+type Validator interface {
+	Validate() error
+}
+
+var (
+	_ Validator = Configuration{}
+	_ Validator = ProviderConfig{}
+	_ Validator = SessionConfig{}
+	_ Validator = CookieConfig{}
+	_ Validator = TTLConfig{}
+	_ Validator = ClientConfig{}
+	_ Validator = ServerConfig{}
+	_ Validator = TimeoutConfig{}
+	_ Validator = MetricsConfig{}
+	_ Validator = StatsdConfig{}
+	_ Validator = LoggingConfig{}
+	_ Validator = UpstreamConfigs{}
+	_ Validator = DefaultConfig{}
+	_ Validator = EmailConfig{}
+	_ Validator = RequestSignerConfig{}
+)
+
+type Configuration struct {
+	ServerConfig        ServerConfig        `mapstructure:"server"`
+	ProviderConfig      ProviderConfig      `mapstructure:"provider"`
+	ClientConfig        ClientConfig        `mapstructure:"client"`
+	SessionConfig       SessionConfig       `mapstructure:"session"`
+	UpstreamConfigs     UpstreamConfigs     `mapstructure:"upstream"`
+	MetricsConfig       MetricsConfig       `mapstructure:"metrics"`
+	LoggingConfig       LoggingConfig       `mapstructure:"logging"`
+	RequestSignerConfig RequestSignerConfig `mapstructure:"requestsigner"`
+}
+
+func (c Configuration) Validate() error {
+	var validationErrors []string
+
+	if err := c.ServerConfig.Validate(); err != nil {
+		validationErrors = append(validationErrors, fmt.Sprintf("invalid server config: %s", err))
+	}
+
+	if err := c.ProviderConfig.Validate(); err != nil {
+		validationErrors = append(validationErrors, fmt.Sprintf("invalid provider config: %s", err))
+	}
+
+	if err := c.SessionConfig.Validate(); err != nil {
+		validationErrors = append(validationErrors, fmt.Sprintf("invalid session config: %s", err))
+	}
+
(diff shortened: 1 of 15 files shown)


internal/proxy/configuration.go
```diff
@@ -0,0 +1,462 @@
...
+// UPSTREAM_DEFAULT_TCP_RESET_DEADLINE
+// UPSTREAM_DEFAULT_PROVIDER_SLUG
+// UPSTREAM_CONFIGS_FILE
+// UPSTREAM_SCHEME
+// UPSTREAM_CLUSTER
+//
+// PROVIDER_TYPE
+// PROVIDER_URL_EXTERNAL
+// PROVIDER_URL_INTERNAL
+// PROVIDER_SLUG
```
Jusshersmith, line 58: s/provider_slug/provider

Jusshersmith, line 58: 👀

internal/proxy/configuration.go
```diff
@@ -0,0 +1,462 @@
...
+//
+// LOGGING_ENABLE
+// LOGGING_LEVEL
+//
+// UPSTREAM_DEFAULT_EMAIL_DOMAINS
+// UPSTREAM_DEFAULT_EMAIL_ADDRESSES
+// UPSTREAM_DEFAULT_GROUPS
+// UPSTREAM_DEFAULT_TIMEOUT
+// UPSTREAM_DEFAULT_TCP_RESET_DEADLINE
+// UPSTREAM_DEFAULT_PROVIDER_SLUG
```
Jusshersmith, line 50: 👀

Jusshersmith:
Any thoughts on the naming of config variables themselves would be appreciated. I've tried to migrate all config variables while maintaining a tidy, logical structure..but that's not always straight forward.

internal/proxy/options.go
```diff
@@ -1,311 +1,197 @@
...
+func SetProvider(provider providers.Provider) func(*OAuthProxy) error {
+	return func(op *OAuthProxy) error {
+		op.provider = provider
+		return nil
 	}
+}
 
-	if o.UpstreamConfigsFile != "" {
-		rawBytes, err := ioutil.ReadFile(o.UpstreamConfigsFile)
+func SetUpstreamConfigs(uc *UpstreamConfigs, cc *CookieConfig, svc ServerConfig) error {
```
Integralist, line 107: Looks like `cc` doesn't need to be a pointer as we're only reading a value.

So if you make that change you'll also need to pass `config.SessionConfig.CookieConfig` as a value.

internal/proxy/configuration.go
```diff
@@ -0,0 +1,458 @@
...
+// SERVER_PORT
+// SERVER_TIMEOUT_SHUTDOWN
+// SERVER_TIMEOUT_READ
+// SERVER_TIMEOUT_WRITE
+//
+// METRICS_STATSD_HOST
+// METRICS_STATSD_PORT
+//
+// LOGGING_ENABLE
+// LOGGING_LEVEL
```
Jusshersmith, line 43: comment needs to be removed. This config var was included initially, but didn't seem to actually be used anywhere.

internal/proxy/configuration.go
```diff
@@ -0,0 +1,458 @@
...
+// UPSTREAM_CLUSTER
+//
+// PROVIDER_TYPE
+// PROVIDER_URL_EXTERNAL
+// PROVIDER_URL_INTERNAL
+// PROVIDER_SCOPE
+
+func DefaultProxyConfig() Configuration {
+	return Configuration{
+		ServerConfig: ServerConfig{
```
Integralist, line 62: Considering the name of this package is already called `configuration` and it's pulled in/referenced elsewhere as `config` I don't believe we need individual fields to also include 'config' in their name.

It reads fine in most cases, e.g. `config.LoggingConfig` actually kinda _sounds_ better than `config.Logging` but strictly speaking it feels like it should be redundant.

But on the flip-side that 'sounding better' argument gets weaker when we get to examples like `config.ServerConfig.TimeoutConfig` vs `config.Server.Timeout`.

internal/proxy/configuration.go
```diff
@@ -0,0 +1,458 @@
...
+	"encoding/base64"
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+
+	"github.com/micro/go-micro/config"
+	"github.com/micro/go-micro/config/source/env"
+	"github.com/mitchellh/mapstructure"
+	"golang.org/x/xerrors"
```
Integralist, line 13: My understanding is that we are using this simply to provide users with support for `Is`, `As`, `Unwrap` and `%w` in all 1.x go releases (whatever version of go they happen to be using). Is that correct?

internal/proxy/configuration.go
```diff
@@ -0,0 +1,458 @@
...
+	if err := c.ProviderConfig.Validate(); err != nil {
+		return xerrors.Errorf("invalid provider config: %w", err)
+	}
+
+	if err := c.SessionConfig.Validate(); err != nil {
+		return xerrors.Errorf("invalid session config: %w", err)
+	}
+
+	if err := c.ClientConfig.Validate(); err != nil {
+		return xerrors.Errorf("invalid client config: %w", err)
```
Integralist, line 153: should we add tests for these functions?

Integralist (dismissed):
I'm going to approve although this was difficult for me to review.

I would personally have preferred mulitple PRs for the sake of the various refactoring work (e.g. one PR to move loads of code from one file to another, then a separate PR for the cookie renaming etc). That way focusing in on the actual meat of the changes would be easier.

That said I appreciated the effort put into the README.

internal/proxy/configuration.go
Jusshersmith: SSO is built from Go 1.14 now, so this is actually probably unnecessary. I believe this pull request was originally written prior to that update when we were on 1.12. Though I think I'd prefer in this case to keep it as is in this pull request, and return with a PR to refactor this in both places (`sso_auth` and `sso_proxy`) together. Let me know if you have any objections there!

internal/proxy/options.go
Jusshersmith: Good catch, thank you!

internal/proxy/configuration.go
Jusshersmith: Individual configuration structs have their `Validate()` method tested in the `TestConfigValidate` function [here](https://github.com/buzzfeed/sso/pull/279/files#diff-b2162c167e6567cc0326346abf552416R101). Though I've also refactored this catch-all `Configuration.Validate()` method slightly to not return after the first error, but continue until all configurations have been tested and return all errors together.

This made it easier to add in a test to validate the functionality within this method (`Configuration.Validate()`) itself.

internal/proxy/configuration.go
Jusshersmith: This pull request was originally written prior to SSO's update to Go 1.14, with it previously being on Go 1.12. I've moved this over to use the default `fmt` library (I think the main motivation here was to enable the use of `%w`, though now that's supported as of Go 1.13).

I'll look to update this also in `internal/auth/configuration.go` to realign the two - there are also a couple other changes to make).

Jusshersmith: This pull request was originally written before SSO's move to Go 1.14 (previous to that it was on Go 1.12). I believe the main initial motivation was to enable use of `%w`, though now that's supported (as of Go 1.13) I think we can move to use the default `fmt` and `errors` libraries.

I'll look to open a subsequent pull request to make the same change in `sso_auth` to realign the two on this. (there are also a few other similar changes that need to be made)

internal/proxy/configuration.go
Jusshersmith: Interesting. I think in most cases you're right, though because they are their own independent, separate structs I'm not sure how well it would read in the receiving functions, e.g:
`func SetCookieStore(cc CookieConfig) func(*OAuthProxy) error {` 
vs
`func SetCookieStore(cc Cookie) func(*OAuthProxy) error {` 
or in some tests. I need to have a look closer look through.

I think this is a good point and definitely worth considering, though I think I'd personally opt to look at this in a separate followup pull request - and look at this and `sso_auth`'s equivalent configuration package _together_ in this aspect.

Let me know if you have any objections there!

benjsto (approved):

PR #284, opened 2020-03-06
Proxy path-components with %-escaped characters in tact.

katzdm:
## Problem

When proxying to a path with a %-encoded `/` character (i.e. `%2F`), the Golang `http.ServeMux` class auto-unwraps the %-encoding. It then uses `path.Clean()` to "helpfully" normalize successive `/` characters (e.g. `/a/b//c` to `/a/b/c`, `/a/b/../c` to `/a`, etc).

Though admittedly an edge-case, the unintended side-effect is that a URL whose path contains a %-encoded URL will be proxied incorrectly. Fo instance, the URL
`https://example.com/path/http:%2F%2Ffoo.com/`
will be proxied to
`https://example.com/path/http:/foo.com/`.

## Solution

Replace use of `http.ServeMux` with the `mux.Router` class from the popular https://github.com/gorilla/mux library, which allows use of `URL.EscapedPath()` in lieu of directly reading `URL.Path`. This preserves the %-wrapping of path-components, which in turn prevents `path.Clean()` from errantly rewriting the path.

## Notes

The motivating example derives from the popular open-source Jenkins project, which uses URLs in such a form to check the health of a reverse-proxy - Hence this bug causes a Jenkins instance behind an SSO deployment to report a "broken" reverse-proxy configuration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,8 @@ ENV GO111MODULE=on
 
 WORKDIR /go/src/github.com/buzzfeed/sso
 
+COPY ./go.mod ./go.sum ./
+RUN go mod download
 COPY . .
 RUN cd cmd/sso-auth && go build -mod=readonly -o /bin/sso-auth
 RUN cd cmd/sso-proxy && go build -mod=readonly -o /bin/sso-proxy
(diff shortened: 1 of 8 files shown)


katzdm:
> This looks good -- but I'd like to see a test added to the proxy side.

Great idea - Done; I've added one to `oauthproxy_test.go`.

PR #296, opened 2020-05-30
Updated dockerfile to golang 1.14

ducksecops:
## Problem

Golang 1.12 is now EOL.
Can't locate Term/ReadLine.pm in @INC in docker build.

## Solution

Update dockerfile to use golang:1.14 as build image.
Added "ENV DEBIAN_FRONTEND noninteractive" to dockerfile

## Notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 defaults: &defaults
   docker:
-    - image: circleci/golang:1.12
+    - image: circleci/golang:1.14
   working_directory: /go/src/github.com/buzzfeed/sso
 
 attach_workspace: &attach_workspace
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 #
 # install golang dependencies & build binaries
 # =============================================================================
-FROM golang:1.12 AS build
+FROM golang:1.14 AS build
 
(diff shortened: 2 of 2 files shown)


Jusshersmith:
Hi @ducksecops! Thank you for opening this pull request!

A couple things:
- I think you'll also need to update the go version in `go.mod` to `1.14.` You can do this using the `go mod edit -go=1.14` command. (see the current build failure)
- I'm not sure it's best to set `DEBIAN_FRONTEND` to `noninteractive` as an environment variable in the Dockerfile. This setting persists after build, and can cause issues if anyone debugging in the container ever _does_ open an interactive shell. 

Is the lack of `noninteractive` being set causing issues? If needed, I think we could probably just set it like so in the Dockerfile instead:
```
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install.........
```

As far as I'm aware it only throws up a few warning messages if not set (without failing) - but I could be wrong!

I suspect there may also be some other small syntax/logic changes needed as part of the move to go 1.14 (we'll see once the build succeeds). If you don't have the time to look into those - I'd be happy to.

Jusshersmith:
Hi @ducksecops,

To keep this moving forward I've finished off and merged this change here: https://github.com/buzzfeed/sso/pull/298.

Thanks again!

PR #299, opened 2020-07-20
sso_*: prevent copying of session between upstreams

Jusshersmith:
## Problem

_We **_only_** revalidate group membership _**when**_ the session is refreshed or revalidated_, which means that if a user were to successfully authorize with upstream 'foo', then they can effectively skip group membership validation on a _different_ upstream by making the request with a slightly altered version of the saved cookie from the 'foo' upstream (providing the session is still valid and hasn't expired).

## Solution

Add a new `AuthorizedUpstream` value to the session which is used to compare the upstream the session has been validated against, to the requested upstream. 
The `AuthorizedUpstream` value is checked against the request host on each request. For the time being, when caught this check will re-trigger the start of the oauth flow, primarily to help introduce this additional check in a graceful manner.

Example log line when triggered:
```
{"authorized_upstream":"upstream-1.foo.io","level":"warning","msg":"session authorized against different upstream; restarting authentication","proxy_host":"upstream-2.foo.io","remote_address":"x.x.x.x","service":"sso-proxy","time":"2020-01-02 13:26:31.502","user":"foo.bar@email.com"}
```
`upstream-1.foo.io` being the original upstream the session was used against, and `upstream-2.foo.io` being the newly 
 requested upstream.

## Notes

- This change was originally made and approved as part of https://github.com/buzzfeed/sso/pull/275, however that pull request was found to also introduce some bugs, so this smaller change has been pulled out to reduce the complexity there and allow for easier troubleshooting. (the bugs are _not_ present in this pull request!)


internal/proxy/oauthproxy.go
```diff
@@ -655,6 +663,12 @@ func (p *OAuthProxy) Proxy(rw http.ResponseWriter, req *http.Request) {
 			// the user has a stale sesssion.
 			p.OAuthStart(rw, req, tags)
 			return
+		case ErrUnauthorizedUpstreamRequested:
+			// The users session has been authorised for use with a different upstream than the one
+			// that is being requested, so we trigger the start of the oauth flow.
+			// This exists primarily to implement some form of grace period while this additional session
+			// check is being introduced.
+			p.OAuthStart(rw, req, tags)
```
Integralist, line 671: Question: will this invalidate the current validated session for 'foo' upstream?

Jusshersmith: It doesn't invalidate the session on the original upstream - and because it restarts the oauth flow for the _new_ request/upstream the UX should be pretty seamless when this is triggered.

PR #301, opened 2020-10-19
sso_proxy: document provider slug configs

Jusshersmith:
## Problem

Some documentation for the upstream configuration option `provider_slug`, and the global `sso_proxy` 
`DefaultProviderSlug` option is missing

## Solution
Add in the missing documentation.


jphines:
This is annoying. Do we want to force merge?

Jusshersmith:
@jphines Actually, I think that might be a good idea. I've tried to dig into why it's suggesting this change in code coverage, but unclear at the moment. 

If you're happy with the change, let's force merge 👍

PR #312, opened 2021-06-16
sso_*: rename 'options' package to 'validators'

Jusshersmith:
## Problem
In `internal/pkg/options` we offer a set of tools to help validate sessions. The name of the package, 'options', doesn't fit its responsibility very well, which can contribute to unnecessary confusion while reading through the code.
## Solution
Separated out from some other cleanups, rename the `options` package to `validators`, and adjust any references to it.

## Notes


benjsto (approved):
looks good to me

PR #314, opened 2021-06-28
quickstart - rename CLUSTER to UPSTREAM_CLUSTER

mirobertod:
## Problem

quickstart is not fully updated with the new variables name introduced with the v3.0.0

## Solution

rename CLUSTER to UPSTREAM_CLUSTER in sso-proxy
remove CLUSTER from sso-auth. it looks like is not used anymore (https://github.com/buzzfeed/sso/blob/main/internal/auth/configuration.go)


Jusshersmith (approved):
Thanks for noticing and fixing this, @mirobertod!

PR #316, opened 2021-07-02
sso_*: allow cookie samesite configuration

Jusshersmith:
## Problem

The [SameSite Cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) attribute allows you to determine whether a cookie is able to be used in a cross-site context, or if it's restricted to a first-party context. 
- Prior to Chrome version 80, if this attribute wasn't explicitly set it would default to `SameSite=None`, allowing cookies to be used in a cross-site context.
- As of and beyond Chrome 80, the default behavior is now `SameSite=Lax`, restricting cookies to a first-party or same-site context.

This has no impact on typical workflows, however is a requirement for others.

## Solution

~Add a `session_cookie_samesite` configuration variable (`"none"`, `"lax"`, or `"strict"`) to both `sso_proxy` and `sso_auth`, which is then stored in the session cookie store and set on any created cookies.~

The above solution stored the `SameSite` variable in the cookie store and set the value on _all_ created cookies. The new solution (below) allows `sso_proxy` to read the setting from upstream configurations, instead of globally, but it does also cause some less-appealing workflows (like having to pass it explicitly in calls to `SetCSRF`, for example).

### sso_auth
- Add a `session_cookie_samesite` global configuration variable (`"none"`, `"lax"`, or `"strict"`), which is stored in the `Authenticator` object, and subsequently, the session itself. 
### sso_proxy
- Add a `cookie_samesite` upstream configuration variable (`"none"`, `"lax"`, or `"strict"`), which is then stored in the `OAuthProxy` object, and subsequently, the session itself.

### session pkg
- `SameSite` has been added to the session state, which is then referenced in chained calls to set the session cookie.
- ~SetCSRF doesn't make use of the session state, and so now (a little annoyingly) requires a sameSite to be explicitly passed in.~
- `SetCSRF` always sets `http.SameSiteDefaultMode`. 
- (`ClearSession`, and `ClearCSRF` both set `http.SameSiteDefaultMode` -- I don't _think_ (I might be wrong!) this has much of an effect, but this felt like the safest setting?)

## Notes

- If left unset, then `SameSiteDefaultMode` is used. (https://golang.org/src/net/http/cookie.go)
- If `SameSite` is set to `None`, then `Secure` [must be enabled](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#samesitenone_requires_secure).


Jusshersmith:
Although in certain situations having this flexibility may be useful, we've opted, at least for now, not to ship this.

A change like this has potentially wide yet obscure impacts on the security posture of this application, so we would prefer to avoid this route unless absolutely required.

PR #319, opened 2021-07-23
update the makefile to push multi-arch images

danbf:
## Problem

Please describe the problem this PR will solve.

## Solution

What is the proposed solution?

## Notes


danbf:
seems to be working

Screen Shot 2021-07-23 at 12 34 30 AM

danbf:
will test tomorrow to ensure they actually run on the platforms

danbf:
ok got it running in stage and it seems fine, both sso_proxy and sso_auth

Dockerfile
```diff
@@ -24,6 +24,10 @@ RUN cd cmd/sso-proxy/generate-request-signature && go build -mod=readonly -o /bi
 # add static assets and copy binaries from build stage
 # =============================================================================
 FROM debian:stable-slim
+RUN ln -s /usr/bin/dpkg-split /usr/sbin/dpkg-split
+RUN ln -s /usr/bin/dpkg-deb /usr/sbin/dpkg-deb
+RUN ln -s /bin/tar /usr/sbin/tar
+RUN ln -s /bin/rm /usr/sbin/rm
```
danbf, line 30: this was needed for the builds to complete.  seems like the file locations are in different spots on the ARM arch.

Makefile
```diff
@@ -26,16 +26,25 @@ clean:
...
 
 releasepush:
-	docker build -t buzzfeed/sso:$(version) -t buzzfeed/sso-dev:latest .
-	docker push buzzfeed/sso:$(version)
-	docker push buzzfeed/sso:latest
+	docker context create sso-dev
+	docker buildx create sso-dev --name sso-dev
+	docker buildx use sso-dev
+	docker buildx build --tag buzzfeed/sso:$(version) . --platform linux/amd64,linux/arm64,linux/arm/v7 --push
+	docker buildx build --tag buzzfeed/sso-dev:latest . --platform linux/amd64,linux/arm64,linux/arm/v7 --push
```
Jusshersmith, line 47: I think we might also want to push a `buzzfeed/sso:latest` tag. I'm also not particularly sure we really need the `buzzfeed/sso-dev:latest` tag to be honest. That tag should be pushed with each push to the `main` branch (through [this](https://github.com/buzzfeed/sso/blob/main/.circleci/config.yml#L85-L90) logic), so I'm not sure we gain anything with it being here as well.

Dockerfile
Jusshersmith: 👍 Would it be worth adding this in a quick inline comment?

Makefile
danbf: thanks, this was a typo, should have been `buzzfeed/sso:latest`.  fixing it.

Dockerfile
danbf: if this remains, yes.  i want to test the resultant multi-arch images in a arm64 cluster before finalizing this.  i only got to test it once during hackweek before tearing down the temp cluster.

danbf: i'm going to see if we need this for arm64, which is the primary secondary target we want for M1's and Graviton's

danbf: that didn't work, these are needed for arm64 also apparently.

PR #322, opened 2021-08-05
sessions pkg: minor refactors

Jusshersmith:
## Problem
In an effort to reduce the scale of some other changes I'm trying to pull through, we make some small refactors here to `internal/proxy/providers/sso.go`.

## Solution
Included refactors are:
- Pull out the `withinGracePeriod` method within `internal/proxy/providers/sso.go` into a new method within the sessions package: `IsWithinGracePeriod`.
- Instead of calling a local `extendDeadline` function within `internal/proxy/providers/sso.go`, call the already existing sessions package `ExtendDeadline` method.

### DIFF
--- a/internal/pkg/sessions/session_state.go
+++ b/internal/pkg/sessions/session_state.go
@@ -53,6 +53,14 @@ func isExpired(t time.Time) bool {
 	return false
 }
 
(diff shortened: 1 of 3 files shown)


laurenzhang (approved):
these changes lgtm!

PR #326, opened 2021-09-11
updating k8s configuration

2o1o0:
## Problem

Kubernetes configuration file are out of date

## Solution

Match with current sso and kubernetes requirements to make it work

## Notes

As I followed the original article to setup buzzfeed/sso on K8S back in the days, I kind of lost tracked of what env vars are actually required. Maybe some I use are now out of date but are still available on my cluster.


RobertLowe:
demo-apps also need updating

2o1o0:
@RobertLowe As this project didn't seems to be updated anymore, I switched to keycloak + Oauth2 proxy to achieve the same, but with what I find to be a better integration overall.

PR #329, opened 2021-12-15
sessions pkg: omit domain attribute in session cookie

Jusshersmith:
## Problem
By default, we set the `Domain` attribute in the session cookie to the request's host value. This causes the cookie to be valid for the provided domain, **and all subdomains**. I don't believe this is expected behaviour, and shouldn't be the default. 

For reference: https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.3
> The Domain attribute specifies those hosts to which the cookie will
   be sent. For example, if the value of the Domain attribute is
   "example.com", the user agent will include the cookie in the Cookie
   header when making HTTP requests to example.com, www.example.com, and
   www.corp.example.com.
> ...
> If the server omits the Domain attribute, the user
   agent will return the cookie only to the origin server.

A leading `.` no longer has any impact on where the cookie is shared.

More references: [mozilla domain attribute](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#:~:text=be%20sent%20to.-,Domain%20attribute,-The%20Domain%20attribute), [mozilla set-cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#:~:text=Age%20has%20precedence.-,Domain%3D%3Cdomain%2Dvalue%3E,-Optional)

## Solution

Unless a domain is explicitly specified in the configuration, set the domain value to `""` so that the cookie isn't shared with subdomains. Instead, it automatically defaults to _only_ the request's domain

### DIFF
--- a/internal/pkg/sessions/cookie_store.go
+++ b/internal/pkg/sessions/cookie_store.go
@@ -3,7 +3,6 @@ package sessions
 import (
 	"errors"
 	"fmt"
-	"net"
 	"net/http"
 	"strings"
 	"time"
@@ -81,13 +80,14 @@ func NewCookieStore(cookieName string, optFuncs ...func(*CookieStore) error) (*C
 
 func (s *CookieStore) makeCookie(req *http.Request, name string, value string, expiration time.Duration, now time.Time) *http.Cookie {
 	logger := log.NewLogEntry()
-	domain := req.Host
-	if h, _, err := net.SplitHostPort(domain); err == nil {
-		domain = h
-	}
+
+	// To avoid subdomains being inadvertently included, omit the domain attribute unless
+	// it's explicitly set in s.CookieDomain.
+	// https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.3
+	domain := ""
 	if s.CookieDomain != "" {
 		if !strings.HasSuffix(domain, s.CookieDomain) {
-			logger.WithRequestHost(domain).WithCookieDomain(s.CookieDomain).Warn("Warning: Using configured cookie domain.")
+			logger.WithRequestHost(domain).WithCookieDomain(s.CookieDomain).Warn("Warning: Using explicitly configured cookie domain.")
 		}
 		domain = s.CookieDomain
 	}
--- a/internal/pkg/sessions/cookie_store.go
+++ b/internal/pkg/sessions/cookie_store.go
@@ -3,6 +3,7 @@ package sessions
 import (
 	"errors"
 	"fmt"
+	"net"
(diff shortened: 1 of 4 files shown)


internal/pkg/sessions/cookie_store.go
```diff
@@ -81,13 +80,14 @@ func NewCookieStore(cookieName string, optFuncs ...func(*CookieStore) error) (*C
...
-	domain := req.Host
-	if h, _, err := net.SplitHostPort(domain); err == nil {
-		domain = h
-	}
+
+	// To avoid subdomains being inadvertently included, omit the domain attribute unless
+	// it's explicitly set in s.CookieDomain.
+	// https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.3
+	domain := ""
 	if s.CookieDomain != "" {
```
benjsto, line 88: I'm not actually finding any cases (other than test code) where `CookieDomain` (via the `CookieConfig.Domain` value) would be set -- am I missing something?

benjsto (approved):

internal/pkg/sessions/cookie_store.go
Jusshersmith: Right, as far as I understand the only other time it's set is if it's explicitly set as a configuration variable.

As you've mentioned, that would be through the `CookieConfig.Domain` configuration variable, which is then set as part of the `CookieStore` here (and the equivalent in `sso_auth`):
https://github.com/buzzfeed/sso/blob/9c912b1e1d6e3cc235795a69240ce167c27954ad/internal/proxy/options.go#L30-L38


I suppose we could simply set the cookie domain attribute to always match the `CookieConfig.Domain` value, seeing as it will either be explicitly set, in which case we want to use it, or the zero value of `""` anyway.

benjsto: We're on the same page. I think what you've got now makes a lot of sense but maybe ought to be documented over [here](https://github.com/buzzfeed/sso/blob/main/docs/sso_authenticator_config.md#L14), to make very clear the behavior of omitting that configuration value or setting it empty?

Jusshersmith: I made a small adjustment to the documentation there already, but I think you're right -- this cookie behavior (even outside of this application) isn't necessarily what one might expect, so making that even more explicit and clear is 👍

Jusshersmith: I've pushed up a change with some additional changes to the documentation.

benjsto (approved):

PR #335, opened 2022-08-29
pkg/logging: Use RFC3339 compatible time format

namm2:
Signed-off-by: Nam Hai Nguyen <namhai.nguyen-ext@commercetools.com>

## Problem

Logging pkg doesn't use either RFC3339 or ISO 8601 time format, then we have to parse SSO logs specifically.

## Solution
Use a valid RFC 3339 time format.

Referece: https://ijmacd.github.io/rfc3339-iso8601/

## Notes

### DIFF
--- a/internal/pkg/logging/logging.go
+++ b/internal/pkg/logging/logging.go
@@ -12,7 +12,7 @@ var serviceName = "sso"
 
 func init() {
 	logrus.SetOutput(os.Stdout)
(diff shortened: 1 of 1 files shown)


svenmueller:
How can we get a PR like this reviewed and merged? @sporkmonger @mcfearsome @joelsouza @benjsto

PR #352, opened 2024-12-29
fix(deps): Integer overflow in websocket 

kreeksec:
## Problem
An integer overflow vulnerability exists with the length of websocket frames received via a websocket connection. An attacker would use this flaw to cause a denial of service attack on an HTTP Server allowing websocket connections.


danbf:
handling in https://github.com/buzzfeed/sso/pull/353 since newer package versions have come out

